Issue #12, opened 2026-04-30
install.sh leaves stale hook entries when upgrading from a previous install path (e.g. tinynudge → stack-nudge)


OMauriStkOne:
## Problem

When `install.sh` runs on a machine that already has hook entries pointing at an older install path (e.g. the legacy `~/.tinynudge/notify.sh` from before the rename, or any prior `~/.stack-nudge/notify.sh` entry that's been moved/edited), the installer **appends** new entries instead of **replacing** the existing ones.

After upgrade, `~/.claude/settings.json` ends up with two `Stop` hook blocks and two `PermissionRequest` hook blocks: one stale, one fresh. Both fire on every event. The stale one points at a path that no longer exists once the user removes the old install, so every Stop and PermissionRequest then logs a "no such file or directory" error.

## Repro

1. Have an older config wired up:
   ```json
   "Stop": [
     { "matcher": "", "hooks": [{ "type": "command", "command": "/Users/me/.tinynudge/notify.sh claude-code stop" }] }
   ]
   ```
2. Run `./install.sh` from the current repo.
3. Inspect `~/.claude/settings.json` — the new `~/.stack-nudge/notify.sh` block has been appended alongside the old one rather than replacing it.

Same issue applies to `~/.cursor/hooks.json`.

## Expected

The installer should detect any existing hook entry whose command path looks like a prior stack-nudge / tinynudge `notify.sh` invocation (matching `*/notify.sh claude-code <event>` or similar) and replace it in-place, rather than appending.

## Suggested fix

In the Python snippets at `install.sh:167` and `install.sh:199` that mutate the JSON config, before appending the new hook block, walk the existing `hooks[event]` list and drop any entry whose nested `hooks[*].command` matches a regex like:

```
.*/(tinynudge|stack-nudge)/notify\.sh\b
```

Then append the fresh entry. That keeps non-stack-nudge hooks (e.g. pixel-agents, custom user hooks) untouched while idempotently replacing our own.

## Why this matters

- Silent failures: the dead hook command runs on every Stop/PermissionRequest and exits non-zero.
- Surprises users who don't notice duplicates until they `cat` the file or notice extra noise.
- `./uninstall.sh` likely has the same blind spot — worth a look there too.

## Related

Filed during the same install session as #11.

StuBehan:
[...]

Issue #11, opened 2026-04-30
install.sh fails silently when system Python is < 3.10 (error swallowed by Swift build warnings)


OMauriStkOne:
## Problem

On a fresh macOS install where the only `python3` on PATH is the system Python (3.9.6 on macOS 14/15), `install.sh` exits with status 1 immediately after printing `Setting up voice engine...` with no obvious indication of what went wrong.

The installer *does* print a helpful message (`install.sh:56-57`):

> Could not find Python ≥ 3.10. Install one (e.g. `brew install python@3.13`) or set `STACKNUDGE_PYTHON=/path/to/python3` and re-run.

…but on macOS that message is buried under ~120 lines of `NSUserNotification` Swift deprecation warnings emitted during the `build/stack-nudge.app` step. Anyone scrolling the install output (or using `tail`) misses the actionable line entirely and just sees a non-zero exit.

## Repro

1. macOS with no Homebrew Python and no pyenv (only `/usr/bin/python3` → 3.9.6).
2. `git clone` and `./install.sh`.
3. Build succeeds, app bundles install, then exits 1 after `Setting up voice engine...`.

## Suggested fixes

- Silence or filter the Swift deprecation warnings during `build.sh` (e.g. `-suppress-warnings` or redirect to a log) so install output stays scannable.
- Make the "no Python ≥ 3.10" message louder — e.g. blank lines around it, a `❌` marker, or print it again at the very end before `exit 1`.
- Optionally: detect Homebrew and offer to run `brew install python@3.13` automatically (with a yes/no prompt).
- Bonus: docs note in README install section that Python ≥ 3.10 is a prerequisite.

## Environment

- macOS Darwin 25.3.0
- `/usr/bin/python3` → Python 3.9.6
- Homebrew 5.1.8 installed but no `python@3.x` formula
- No pyenv

PR #16, opened 2026-04-30
fix: migrate to stackvox 0.4.x unified CLI

### LINKED ISSUE #14: notify.sh calls non-existent `stackvox-say` binary — voice silently broken on stackvox ≥ 0.3.0
## Problem

`notify.sh` shells out to `\${VENV}/bin/stackvox-say` to speak notifications, but that binary does not exist in stackvox 0.3.x. The current stackvox CLI is unified — `say` is a subcommand (`stackvox say ...`), not a separate console_script.

The `speak_notification()` function gates on `[[ ! -x "$STACKVOX_SAY" ]] && return`, so when the binary is missing it bails silently. Users with `STACKNUDGE_VOICE=true` get no voice and no error — install looks healthy, daemon is running, models are downloaded, the manual `stackvox speak` command works, but hooks never speak.

## Repro

1. Fresh install on macOS with `./install.sh`.
2. Set in `~/.stack-nudge/config`:
   ```
   STACKNUDGE_VOICE=true
   STACKNUDGE_VOICE_NAME=af_river
   ```
3. Trigger a Claude Code Stop hook (or run `~/.stack-nudge/notify.sh claude-code stop`).
4. Expected: voice speaks. Actual: silence (banner + sound still fire).

Trace shows the bail:

```
+ STACKVOX_SAY=/Users/me/.stack-nudge/venv/bin/stackvox-say
+ speak_notification 'stack nudge is ready for you'
+ [[ ! -x /Users/me/.stack-nudge/venv/bin/stackvox-say ]]
```

And the venv only contains the unified CLI:

```
$ ls ~/.stack-nudge/venv/bin/ | grep -i stack
stackvox
```

```
$ ~/.stack-nudge/venv/bin/stackvox say --help
usage: stackvox say [-h] [--voice VOICE] [--speed SPEED] [--lang LANG] [--file FILE] [--fallback-say] [text]
```

## Suggested fix

In `notify.sh` (around lines 207–223):

```diff
 STACKVOX="${VENV}/bin/stackvox"
-STACKVOX_SAY="${VENV}/bin/stackvox-say"

 speak_notification() {
   [[ "${VOICE_ENABLED}" != "true" ]] && return
-  [[ ! -x "$STACKVOX_SAY" ]] && return
+  [[ ! -x "$STACKVOX" ]] && return
   local text="$1"
   if [[ ! -S "${HOME}/.cache/stackvox/daemon.sock" ]]; then
     nohup "$STACKVOX" serve >/dev/null 2>&1 &
   fi
   local kokoro_lang
   kokoro_lang=$(voice_to_kokoro_lang "$VOICE_NAME")
-  "$STACKVOX_SAY" --voice "${VOICE_NAME}" --lang "${kokoro_lang}" --speed "${VOICE_SPEED}" "${text}" 2>/dev/null &
+  "$STACKVOX" say --voice "${VOICE_NAME}" --lang "${kokoro_lang}" --speed "${VOICE_SPEED}" "${text}" 2>/dev/null &
 }
```

I patched my local copy this way and voice now fires correctly through the hook path.

## Defensive tweak

The silent bail (`[[ ! -x "$STACKVOX_SAY" ]] && return`) is the reason this hid for who-knows-how-long. Worth logging to a log file (or stderr behind a `STACKNUDGE_DEBUG=true` flag) when voice is requested but the binary is missing — same logic for the daemon-socket check.

## Pin or document the version

`install.sh` pins `stackvox>=0.3.0`. Either:
- Pin a known-working range that still ships `stackvox-say`, or
- Update `notify.sh` to call `stackvox say` (recommended, since the unified CLI is the going-forward shape).

## Environment

- macOS Darwin 25.3.0
- stackvox 0.3.1
- Python 3.13.13 (Homebrew)
- daemon running, models downloaded successfully

## Related

Same install session as #11 and #12.

StuBehan:
## Summary

stackvox 0.4.x consolidated its CLI — there is no separate \`stackvox-say\` console_script anymore; speech goes through \`stackvox say <text>\` as a subcommand. \`notify.sh\` and \`Speaker.swift\` both still pointed at the old binary, so on fresh installs (where only \`stackvox\` exists in the venv) the existence guard bailed silently and voice never fired.

Closes #14, reported by @OMauriStkOne.

## Failure mode this fixes

- Install looks healthy, daemon running, models downloaded
- \`stackvox say "..."\` works manually
- \`STACKNUDGE_VOICE=true\` set
- Hook fires → banner + sound play → **voice never speaks, no error surfaced**

## Changes

**\`notify.sh\`**
- Drop \`STACKVOX_SAY\` variable
- \`speak_notification\` invokes \`"\$STACKVOX" say ...\`
- \`voice_phrase_for\` gates on \`STACKVOX\` existence
- New \`nudge_debug()\` helper writes to stderr when \`STACKNUDGE_DEBUG=true\`. Logs the missing-binary and missing-daemon-socket cases that previously failed silently — so the next time something like this happens, users can diagnose with one env var instead of bisecting the script.

**\`panel/Speaker.swift\`**
- Same migration on the Swift side (used by the menu-bar voice toggle preview and the voice-cycle preview in Settings)

## Test plan

- [x] \`stackvox say --voice af_aoede ...\` runs cleanly against an installed venv
- [x] \`notify.sh claude-code stop\` with \`STACKNUDGE_VOICE=true\` completes the voice path without error
- [x] \`STACKNUDGE_DEBUG=true\` surfaces the diagnostic when stackvox is absent
- [x] \`shellcheck -S warning\` and \`bash -n\` clean
- [x] \`make build\` (swiftc path) and \`swift build\` (Package.swift) both succeed

## Out of scope (deliberate)

- The two related issues from the same install session (#11 install.sh silent failure on Python < 3.10, #12 stale hook entries on upgrade) are independent of this — covered by separate PRs.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -48,7 +48,7 @@ find_python() {
 # Install the voice engine (stackvox) from PyPI into an isolated venv.
 echo ""
 echo "Setting up voice engine..."
-STACKVOX_SPEC="stackvox>=0.3.0"
+STACKVOX_SPEC="stackvox>=0.4.0"
 PYTHON=$(find_python)
 if [[ -z "$PYTHON" ]]; then
   echo "  Could not find Python ≥ 3.10. Install one (e.g. 'brew install python@3.13')"
--- a/notify.sh
+++ b/notify.sh
@@ -146,7 +146,7 @@ voice_phrase_for() {
   local event="$1"
   local lang repo
 
-  if [[ -x "$STACKVOX_SAY" ]]; then
+  if [[ -x "$STACKVOX" ]]; then
     lang=$(voice_to_lang "$VOICE_NAME")
     repo=$(repo_name_raw)
   else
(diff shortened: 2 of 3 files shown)


StuBehan:
Confirmed against [stackvox 0.4.0 on PyPI](https://pypi.org/project/stackvox/0.4.0/) — `stackvox say --voice ... --speed ... --lang ...` is unchanged from 0.3.x (0.4.0 only adds a per-user TOML config layer for defaults). Force-pushed to bump `install.sh`'s pin from `stackvox>=0.3.0` to `stackvox>=0.4.0` so we're explicit about the version we've validated against.

PR #37, opened 2026-05-14
feat: quota tracking via /api/oauth/usage

hiskudin:
## Summary

Adds an in-panel **Usage tab** + **threshold-crossing notifications** for Claude Code subscription quotas. Mirrors what `claude /usage` shows in the TUI, but always-visible and proactive — alerts you when you're about to burn through your weekly cap.

### How it gets the data

Claude Code's `/usage` slash command queries `https://api.anthropic.com/api/oauth/usage` and renders the bars itself. We do the same call directly:

1. Read the OAuth token from the macOS Keychain (`Claude Code-credentials`). One-time system prompt grants access.
2. Extract `claudeAiOauth.accessToken` from the JSON blob.
3. `GET /api/oauth/usage` with the documented headers (`Authorization: Bearer …`, `anthropic-beta: oauth-2025-04-20`).
4. Decode the response into a typed `QuotaSnapshot { fiveHour, sevenDay, sevenDayOpus, sevenDaySonnet }`.

No SwiftTerm, no TUI parsing, no `claude` subprocess. The endpoint is what Claude Code's own statusline depends on, so the data is the same Anthropic shows you.

### What lands

| Surface | Behaviour |
|---|---|
| **Usage tab** (Cmd+3, between Sessions and Settings) | Quota bars for each tier the user's plan has. Color-coded: green < 50%, yellow 50–80%, red ≥ 80%. Reset times per tier. |
| **Poller** | 60s while the panel is visible; 5 min when hidden (configurable via `STACKNUDGE_USAGE_POLL_MIN`). |
| **Threshold notifications** | One banner per (tier, period) when utilization crosses the user-configured threshold. State resets when the tier's `resets_at` advances (new period, fresh budget). |
| **Settings → Usage** | New section with two rows: `Quota alerts` (toggle, default On) + `Alert threshold` (cycle: 50% / 70% / 80% / 90% / 95%, default 80%). |
| **Failure modes** | No token / denied keychain / 401 / 429 → empty state in the Usage tab + silent stderr log. No retries on the first failure; recovered automatically on the next poll. |

### Files

- `panel/SessionUsage.swift` (new) — `QuotaSnapshot` types, `QuotaProbe` (Keychain + HTTP), `UsageView` (SwiftUI tab content).
- `panel/Panel.swift` — Adds `.usage` case in `PanelContentView`, `Cmd+3` hotkey, the poller + threshold-evaluation logic.
- `panel/PanelNav.swift` — `.usage` PanelMode, snapshot + lastUpdated fields, new Settings rows for the alert toggle / threshold cycle, row-layout shifts.
- `panel/Settings.swift` — New "Usage" section between Voice and Actions.
- `build.sh` — Registers the new source.

### Trade-offs / known gotchas

- **The endpoint is unofficial** (Anthropic could change or remove it). The same risk applies to `claude /usage`, ClaudeBar, and every statusline tool. Multiple maintained projects rely on it today.
- **macOS-only.** No Linux fallback; the Keychain access path is macOS-specific. Acceptable since stack-nudge is mac-native.
- **Ad-hoc-signed builds re-prompt for keychain on every rebuild.** Resolved when we eventually sign with Developer ID — same path that helps the auto-updater.
- **Single threshold across all tiers** (not per-tier). Intentional simplification for v1; multi-tier thresholds can come later if users ask.

## Test plan

- [ ] Merge → release-please cuts the next version
- [ ] Install the new release; first launch shows a macOS Keychain prompt for `Claude Code-credentials` — click Always Allow
- [ ] Cmd+3 → Usage tab renders bars matching what `claude /usage` shows in your terminal
- [ ] Cmd+4 → Settings → Usage section shows two rows that toggle/cycle correctly and persist to `~/.stack-nudge/config`
- [ ] Lower the threshold to 50% (or whatever's just above your current util) → wait one poll → notification banner fires once
- [ ] Toggle "Quota alerts" Off → no more banners even when threshold is exceeded

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -51,6 +51,7 @@ build_app "$APP" "stack-nudge" \
   panel/Permissions.swift \
   panel/Settings.swift \
   panel/SessionStore.swift \
+  panel/SessionUsage.swift \
   panel/Sessions.swift \
   panel/Phrases.swift \
   panel/UpdateChecker.swift \
--- a/panel/Panel.swift
+++ b/panel/Panel.swift
@@ -565,12 +565,18 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
         }
     }
 
+    // Cached — banner posts can fire frequently in test/edge cases; avoid
+    // allocating a fresh formatter every time.
+    private static let quotaBannerFormatter: RelativeDateTimeFormatter = {
+        let f = RelativeDateTimeFormatter()
+        f.unitsStyle = .full
+        return f
+    }()
+
     private func postQuotaBanner(label: String, percent: Int, resetsAt: Date?) {
         let body: String
         if let resetsAt {
-            let formatter = RelativeDateTimeFormatter()
-            formatter.unitsStyle = .full
-            body = "\(percent)% used. Resets \(formatter.localizedString(for: resetsAt, relativeTo: Date()))."
+            body = "\(percent)% used. Resets \(Self.quotaBannerFormatter.localizedString(for: resetsAt, relativeTo: Date()))."
         } else {
             body = "\(percent)% used."
         }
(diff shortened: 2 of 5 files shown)


StuBehan (approved):
LGTM

PR #51, opened 2026-05-20
fix: Codex + Gemini auto-wire, agent reconciliation banner, banner-flash fix

hiskudin:
Three connected pieces of agent-handling work landing together.

## 1. Codex + Gemini auto-wire

Bootstrap previously only auto-wired Claude Code and Cursor. Codex wasn't in the BootstrapAgent enum at all; Gemini was a deliberate no-op masquerading as supported. README ✅ (experimental) for both overstated reality.

**Codex** ([docs](https://developers.openai.com/codex/hooks)): config at \`~/.codex/hooks.json\`, matcher-group JSON shape identical to Claude Code, same \`Stop\` + \`PermissionRequest\` event names, JSON payload on stdin. Implementation: add \`.codex\` to \`BootstrapAgent\`, route through the renamed \`wireClaudeShapedHooks\` generic writer.

**Gemini CLI** ([docs](https://geminicli.com/docs/hooks/)): config at \`~/.gemini/settings.json\`, matcher-group shape, events renamed — \`AfterAgent\` (turn end) + \`Notification\` (with \`notification_type=ToolPermission\`). **One critical difference**: Gemini's \`timeout\` is in milliseconds (default 60000), not seconds. We multiply by 1000 in the Gemini case so the on-disk value is 30000; otherwise Gemini kills the hook after 30 ms before the shell forks. **Caveat**: Gemini's Notification event is observability-only — banner surfaces the prompt but Allow/Deny has to happen in the terminal.

## 2. Agent reconciliation banner

Three real-world cases silently miss hook setup today:
1. User installs a new agent (Codex, Aider, etc.) AFTER the bootstrap wizard.
2. A StackNudge release adds support for an agent the user already has installed; existing users auto-update + never see the wizard.
3. User manually deletes a hook entry then forgets.

Adds a banner at the top of Settings listing detected-but-unwired agents:

> **Wire up Codex?**
> Detected on this Mac without StackNudge hooks. Set up to start getting banners.
> [ Set up ] [ Not now ]

After Set up, the same slot flashes a green confirmation for ~3 s so users get feedback. Not now persists to \`~/.stack-nudge/dismissed-agents.json\`; dismissed agents reappear only if they leave + re-enter the unwired set (manual edit, then deletion).

Reconciliation runs on app launch + every Settings open, so post-update and post-agent-install scenarios surface naturally without polling.

## 3. Banner-click flash fix

Clicking a notification banner flashed the panel briefly. Two AppKit delegates both fire on banner-click:
- \`applicationShouldHandleReopen\` (added earlier for Dock-icon reopens) → our handler called \`showPanel()\`.
- \`userNotificationCenter(_:didReceive:)\` → handler called \`NSApp.hide(nil)\`.

Defer the reopen-driven \`showPanel\` by 200 ms; let \`didReceive\` veto via a \`bannerActivationUntil\` deadline. Real app-icon reopens see no veto (200 ms latency added — acceptable).

image
image

## Other in this PR

- README polish from the earlier branch life: badges row, StackOne attribution line, non-political Terms of use section, fixed install snippet, expanded license footer.
- \`ui_improvements.md\` removed from repo (moved to workspace docs folder).
- Wizard pre-selects all detected agents (the Gemini exclusion is gone now that it actually wires).
- Uninstall scrubs the four new event keys (\`AfterAgent\`, \`Notification\`) alongside Claude's two.

## Test plan
- [x] Codex stop event from real \`codex exec\` fires a banner (verified locally).
- [x] Gemini AfterAgent fires a banner after timeout-units fix (verified locally).
- [x] Strip Gemini hooks → banner appears → Set up → green flash → hooks re-written (verified).
- [x] Banner click no longer flashes panel (verified).
- [x] Fresh-install wizard wires all 4 detected agents on a Mac with \`~/.codex\`, \`~/.gemini\`, \`~/.claude\`, \`~/.cursor\` (verified by Hisku's machine state).
- [x] Uninstall removes all 4 event keys from all 4 configs cleanly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,11 @@
 |-------|--------|
 | Claude Code | ✅ |
 | Cursor | ✅ |
-| Gemini CLI | ✅ *(experimental)* |
-| Codex | ✅ *(experimental)* |
+| Codex | ✅ |
+| Gemini CLI | ✅ † |
 | Any hooks-capable agent | ✅ — point it at `notify.sh` |
+
+† Gemini's tool-permission hook is observability-only — the banner shows the prompt, but the actual Allow / Deny click has to happen in Gemini's terminal. Claude Code, Cursor, and Codex permission events can be approved from the panel directly.
 
 **Platforms:** macOS — full app with panel, click-to-focus banners, auto-update, quota tracking, voice. Linux (PulseAudio / ALSA / libnotify) and Windows (Git Bash / WSL) get audio + basic notifications via `notify.sh` only.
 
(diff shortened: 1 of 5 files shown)


StuBehan (approved):
LGTM

PR #81, opened 2026-06-09
fix(energy): reduce idle widget rendering and session polling overhead

StuBehan:
## Summary

The compact widget (pill + quota gauge + mascots) animated continuously at ~20 fps and the session poller ran a fixed 3 s loop that shelled out to `lsof`/`ps` many times per scan — burning CPU and energy even when nothing was happening. This PR makes both rendering and polling idle-aware: animations drop to a 60 s heartbeat when idle, session polling backs off to 15 s when the Sessions tab isn't open, and each scan now uses two batched subprocess calls instead of dozens.

## Changes

**Idle-aware widget rendering — `panel/CompactView.swift`**
- Replaced every `TimelineView(.animation(minimumInterval: 0.05))` (continuous ~20 fps redraw) with `TimelineView(.periodic(by:))` driven by a per-view interval.
- The interval is `0.1 s` only while there's actually something to animate (busy session, hover, an active pulse, or ≥75% / syncing quota) and `60 s` otherwise. Applies to the animated pill background, the quota gauge, and all four mascots (robot, cat, eye, ghost).

**Idle-aware session polling — `panel/SessionStore.swift`, `panel/Sessions.swift`, `panel/Panel.swift`**
- Split the single 3 s cadence into foreground (3 s) and background (15 s). `startForegroundPolling()` runs while the Sessions tab is on screen and reverts to the 15 s background cadence on disappear; the pill keeps reading fresh-enough state from the background poll.
- `startPolling(every:)` no-ops when the requested interval is already running, and the stale-sidecar sweep now runs once per process launch instead of on every (re)start.
- Added a re-entrancy guard so a slow `scan()` can't overlap the next timer tick.

**Batched process inspection — `panel/SessionStore.swift`**
- `scan()` previously called `lsof` once per PID and `ps` ~2× per parent-chain hop per session. It now issues a single `lsof -Fpn` for all candidate PIDs (`readCWDs`) and a single `ps` snapshot of the whole process table (`readProcessTable`), then walks each parent chain in memory. Subprocess count per scan drops from O(sessions × hops) to 2.

**Transcript-refresh dedup — `panel/Panel.swift`**
- `refreshAllClaudeStats()` reads multi-MB transcripts off-main and previously fired on every `sessions` change — including when only `ps`'s elapsed-time display string ticked. The subscription now maps to a `TranscriptRefreshKey` (pid, agent, project path, Claude session ID, live status, last-activity) before `removeDuplicates()`, so transcripts are only re-read on meaningful state changes. Hook events and sidecar activity still refresh immediately, so proactive context/threshold alerts stay current.

**Docs — `README.md`**
- Sessions tab note updated to mention the 15 s background cadence for the compact widget.

## Testing

- [ ] `make build` compiles cleanly.
- [ ] `make reload`, then with no agents running, watch Activity Monitor / the Energy tab — idle CPU should drop now that the widget isn't redrawing at 20 fps and the poller ticks every 15 s.
- [ ] Start a `claude` / `codex` / `gemini` / `agy` session: pill + mascots animate smoothly in the busy state, the Sessions tab still updates on the 3 s cadence while open, and context/quota threshold alerts still fire without waiting for a hook event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ When you press `⏎` on a permission event in a VS Code / Cursor terminal pane, 
 
 #### Sessions tab
 
-Live list of running agent processes (`claude`, `gemini`, `codex` — including node-hosted variants like `gemini-cli`). Polls every 3 seconds while visible. Sessions that exit linger for 30s with `ended Ns ago`.
+Live list of running agent processes (`claude`, `gemini`, `codex` — including node-hosted variants like `gemini-cli`). Polls every 3 seconds while visible and every 15 seconds in the background for the compact widget. Sessions that exit linger for 30s with `ended Ns ago`.
 
 For Claude Code sessions specifically, stack-nudge reads `~/.claude/sessions/<pid>.json` (Claude Code's per-process sidecar) to surface live data without waiting for a hook event:
 
--- a/panel/CompactView.swift
+++ b/panel/CompactView.swift
@@ -614,8 +614,19 @@ private struct RobotMascot: View {
         }
     }
 
+    // Three-tier interval keeps the heavy 10fps cadence for actively
+    // changing visuals while still leaving headroom for the idle micro-
+    // loops (head tilt every ~12s, ear flick every ~10s, blink every ~9s,
+    // ghost bob) — those run off TimelineView ticks and would freeze
+    // entirely at a 60s interval. 0.5s = 2fps is plenty for the slow
+    // idle accents and still ~25× cheaper than the original 0.05s.
+    // Two-tier interval. The 0.5s idle cadence is paired with widened blink /
+    // ear-flick / micro-loop windows below so each animation is reliably
+    // sampled at 2fps. Going lower (e.g. 60s) freezes the idle micro-loops
+    // we added in the mascot-reactions release; going faster wastes cycles
+    // when nothing visibly changing depends on a higher framerate.
     private var timelineInterval: TimeInterval {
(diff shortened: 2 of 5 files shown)


hiskudin (approved):

hiskudin:
Pushed two follow-ups on this branch (39b960e):

**Mascot idle micro-loops** — the 60s TimelineView idle fallback froze the per-mascot animations added in v1.15.0 (robot head-tilt every ~12s, cat ear-flick every ~10s, eye blink every ~9s, ghost bob). At a 60s tick, any time-driven `sin(t * …)` motion can only be sampled by aliasing — looks either static or randomly jumpy. Dropped to 0.5s (2fps) and widened the blink / ear-flick / squint *windows* to 0.5–1.5s so each animation is reliably sampled at the slower cadence. Net: visuals preserved, still ~25× cheaper than the original 0.05s always-on.

**Background poll 15s → 10s** — the pill mascot reflects busy/idle via Claude's sidecar `liveStatus` on this cadence. 15s started to feel like a noticeable delay for "agent just went busy" transitions, especially since the rest of the panel surfaces hook-driven events instantly via the socket. 10s keeps the perceived latency under "feels live" while still preserving ~91% of the steady-state subprocess load reduction (vs the original 3s × ~22-fork design).

Tested locally on top of your batched `discover()` — pill animations look correct in idle/busy/event states; ear-flick + blink visible during normal idle.

PR #103, opened 2026-06-17
fix: harden phrase pipeline, validate IPC fifo_path, fix 'O'-key orphan

StuBehan:
## Summary

Resolves all five HIGH findings from the third `/deep-dive` on v1.19.1. Three are an arbitrary-code-execution class — the phrase pipeline read user-writable files under `~/.stack-nudge/` by *executing* them (the same class PR #98 closed for the config file); the other two harden the permission IPC path and the open-editor gesture.

## Changes

**Phrase pipeline — stop executing user-writable files (RCE)**
- `notify.sh`: parse the `TEMPLATES_*` array literals out of `phrases/<lang>.sh` as **data** via a pure awk text scan, instead of `source`-ing the file.
- `notify.sh`: apply `phrases.user.json` (disabled removals + custom additions) without `eval` — each pool's items go through a function as positional args and come back NUL-delimited, so a custom phrase containing `$(…)` or a quote stays literal.
- `Phrases.swift`: parse the array literals in Swift instead of `bash -c "source <path>"` in the resident, entitlement-holding panel process.

**Permission IPC / gesture**
- `EventListener.swift`: validate the socket-supplied `fifo_path` (`/stack-nudge-perm.` prefix + basename `fifo` + `S_ISFIFO`) before accepting it — a same-uid process could otherwise have the panel write `allow`/`deny` into an arbitrary file or a fabricated pipe. Drop just the path on failure (the nudge still shows).
- `Panel.swift`: `actOnSelected` no longer removes the event before the FIFO check, so `'O'` ("Open editor") on a blocking permission keeps it in the panel (resolvable via Dismiss → deny) instead of orphaning the hook for ~550s.

## Testing

- `swiftc -typecheck panel/*.swift shared/*.swift` — 0 errors.
- Full `./build.sh` (arm64) — clean compile + link + ad-hoc sign.
- `bash -n` + `shellcheck -S warning` clean on `notify.sh`.
- Functional tests of **both** new phrase parsers against a hostile phrase file + `phrases.user.json`: correct extraction, disable/custom behaviour preserved, and `$(…)`/quote payloads proven inert (no code executed). The Swift parser is also verified on a `")"`-inside-a-phrase case. Bash side deliberately avoids 3.2-incompatible constructs (no namerefs / `mapfile -d`).
- Unit tests not run locally (XCTest needs full Xcode); CI runs `swift test`.

### DIFF
--- a/notify.sh
+++ b/notify.sh
@@ -189,42 +189,75 @@ voice_phrase_for() {
     return
   fi
 
-  # shellcheck disable=SC1090
-  source "$lang_file"
+  # Parse the phrase arrays out of the file as DATA — never `source` it. The
+  # file lives in a user-writable install dir, so sourcing ran arbitrary shell
+  # on every hook event. extract_phrase_array does a pure text scan of the
+  # `NAME=( "…" "…" )` literal (handles \" and \\; a `)` inside a quote is safe).
+  extract_phrase_array() {
+    awk -v name="$2" '
+      BEGIN { active=0; instr=0; esc=0; cur=""; opener=name "=(" }
+      {
+        line=$0; i=1; n=length(line)
+        if (!active) {
+          if (substr(line,1,length(opener))==opener) { active=1; i=length(opener)+1 }
+          else next
+        }
+        for (; i<=n; i++) {
+          c=substr(line,i,1)
+          if (instr) {
+            if (esc) { cur=cur c; esc=0 }
+            else if (c=="\\") { esc=1 }
+            else if (c=="\"") { printf "%s%c", cur, 0; cur=""; instr=0 }
+            else { cur=cur c }
+          } else if (c=="\"") { instr=1 }
+          else if (c==")") { active=0; exit }
+        }
+      }
+    ' "$1"
+  }
+  local item
+  TEMPLATES_RESPONSE=()
+  TEMPLATES_NOTIFICATION=()
+  while IFS= read -r -d '' item; do TEMPLATES_RESPONSE+=("$item"); done \
+    < <(extract_phrase_array "$lang_file" TEMPLATES_RESPONSE)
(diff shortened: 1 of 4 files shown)


hiskudin (commented):
All three fixes look right — nothing to block on. Two observations for possible follow-up, both preventive, neither would stop me from shipping this PR:

**1. FIFO validation — parent anchoring + TOCTOU.** The substring check `path.contains("/stack-nudge-perm.")` doesn't require that segment to originate from the trusted parent directory. A same-uid attacker can create `/tmp/attacker/stack-nudge-perm.x/fifo` as a real FIFO and it'll pass the `S_ISFIFO` check. Practical exploit surface is thin (the panel only writes `allow\n`/`deny\n` to the FIFO, and the attacker is essentially receiving data they explicitly asked to receive at their own FIFO), but there's still a TOCTOU between the `stat()` here and the eventual `open()`+write in `writeFIFO` — an attacker could swap the FIFO for a symlink between the two. Cheap defense-in-depth:
- Anchor the parent, e.g. `path.hasPrefix("/var/folders/.../stack-nudge-perm.")` with the actual known TMPDIR-based parent directory, rather than a `contains` check.
- `O_NOFOLLOW` on the `open()` in `writeFIFO` to close the symlink-swap window.

**2. Phrase parser — `)` outside a string terminates the array.** Both the `notify.sh` awk parser and the Swift `parseBashArray` treat any unescaped `)` outside a quoted string as end-of-array. Fine for today's shipped `phrases/*.sh` (`NAME=("…" "…")` with no in-array comments), but if a future language file ever adds `#` comment lines inside the literal and one contains a `)`, the parse silently truncates. Trivial to harden later — skip `#…\n` at the outer state — but easy to forget until it bites.

The `'O'`-key orphan fix and the `Phrases.swift` `Process`-based `bash -c source` removal are exactly right. Nice work.

hiskudin (approved):

PR #104, opened 2026-06-17
feat(widget): mini Usage widget + per-session mute + question-vs-approval split

hiskudin:
## Summary

- **Mini Usage widget**: new pill mode showing only the gauge + reset countdown (no mascot/headline/sessions). Toggle via Settings → Widget → Widget type → Usage. Gauge cluster gains a hover scale+brightness pop, native `.help()` tooltip, and a crossfade legend (`5h xx%` / `7d yy%`) replacing the countdown on hover so "which ring is what" is discoverable without UI clutter.
- **Disable Sequoia drag-to-tile** on the pill by dropping `.resizable` from the panel styleMask in compact mode — also kills the invisible edge resize handles. Restored on full-panel.
- **Per-session mute (`M` in Sessions tab)**: silence banner + sound + voice for events from a specific session; pill ripple and mascot reaction still fire. Persistence reuses `SessionPersistence.key(for:)` (agent::projectPath[::tabId]) so mute survives rename, restart, and PID churn. Stored at `~/.stack-nudge/muted-sessions.json`. Muted rows show a `speaker.slash.fill` icon.
- **Question vs. approval split for Claude Code**: `notify.sh` now inspects `tool_name`; `AskUserQuestion` events skip `has_action_button` and the FIFO so panel Enter focuses the editor instead of writing `allow` (which CC was treating as "press Enter on the highlighted multi-select option"). Codex/Antigravity are unaffected — neither currently surfaces an analogous question event via a hook we subscribe to.
- **M shortcut consolidation**: removed the redundant `M → Compact widget` shortcut + footer hint from Events/Usage/Tickets tabs and dropped the now-unused `enterCompactMode` helper. `M` now means **Mute** exclusively (Sessions tab only).

## Test plan

- [ ] Settings → Widget → Widget type toggles between Sessions (default) and Usage; pill resizes accordingly.
- [ ] In Usage mode, hover the gauge: scale pop + countdown crossfades to two-line legend `5h xx%` / `7d yy%`.
- [ ] Drag the pill toward a screen edge: macOS Sequoia tile-to-half does not engage.
- [ ] Sessions tab → select a row → press `M`: row shows speaker-slash icon. Trigger a notification from that session — no banner/sound/voice, but pill ripples and mascot reacts.
- [ ] Press `M` again to unmute; verify events fire normally.
- [ ] Restart app → mute persists across restart, rename, and PID churn.
- [ ] In Claude Code, run a flow that triggers an `AskUserQuestion` (multi-select). Panel shows the event without an Approve button; Enter raises the terminal instead of selecting an option.
- [ ] In Claude Code, trigger a `Bash` permission: Approve button still present; Enter still writes `allow`.
- [ ] Events / Usage / Tickets tabs: no `Compact [M]` footer hint; `M` is a no-op there.


panel/PanelNav.swift
```diff
@@ -482,6 +495,9 @@ final class PanelNav: ObservableObject {
     // event types we should wire.
     @Published var unwiredAgents:    [BootstrapAgent] = []
     @Published var dismissedAgents:  Set<String>      = []
+    // Mascot reaction + pill ripple stay unmuted by design — only banner /
+    // sound / voice are suppressed for events from these sessions.
+    @Published var mutedSessions:    Set<String>      = []
```
StuBehan: **`mutedSessions` has no eviction.** Keys (`agent::projectPath::tabId`) accumulate for every session ever muted and are never pruned, so `muted-sessions.json` grows monotonically across restarts. Unlike `dismissedAgents` (a tiny fixed set), project/tab keys are effectively unbounded over time. Not urgent — but consider dropping keys with no matching live session in `saveMutedSessions()` (or pruning on load) to keep it bounded.

hiskudin: Good catch — addressed in 9ac68a9 by folding mute into `SessionPersistence` instead of running a parallel store. `mutedSessions: Set<String>` and `muted-sessions.json` are gone; `SessionEntry` gained `muted: Bool` alongside `customName`, so both per-session prefs share one keyspace, one file, and one future eviction story via `lastSeenAt`. Toggling mute off on a session with no custom name removes the entry outright — un-touched sessions stay zero-footprint, matching the file's existing design rule. One-shot migration drains any existing `muted-sessions.json` on first load.

hiskudin: Resolved in 9ac68a9, mute is now part of the session state as opposed to putting it in a store.

StuBehan (approved):
LGTM

PR #106, opened 2026-06-18
fix(panel): honor explicit Quit; align side text with gauge center

hiskudin:
## Summary

Two unrelated post-1.20.0 fixes bundled together.

- **Quit-then-relaunch**: the panel's launchd plist shipped with `KeepAlive: true`, so clicking Quit just made launchd respawn it immediately. Switch to the dict form `KeepAlive: { SuccessfulExit: false }` — restart on crash, but honor clean exits. Applies to new installs (`install.sh` + `Bootstrap.writePlist`) and to existing installs via a one-shot migration in `cleanupPostUpdateBackup` that rewrites the plist on launch and reloads launchd.
- **Side text misalignment**: the resting countdown / hover legend in the pill sat above the gauge digit's y-center because its intrinsic ZStack height differed from the gauge cluster's. Pin `sideText.frame(height: size + 2)` to match the halo Circle so the centering computations agree in both Sessions and Usage pill modes.

## Test plan

- [ ] Fresh install via `install.sh` → click Quit panel from Settings. Process should exit and stay gone.
- [ ] On an existing 1.20.0 install, launch the updated bundle. `~/Library/LaunchAgents/com.stackonehq.stack-nudge.plist` `KeepAlive` should be a `<dict>` with `SuccessfulExit:false`. Quit → does not relaunch.
- [ ] Pill in Sessions mode: gauge digit and the headline text/countdown share the same y.
- [ ] Pill in Usage mode (`Settings → Widget → Widget type → Usage`): gauge digit and the `4h36m`-style countdown share the same y at rest.
- [ ] Hover the mini gauge: gauge digit and the two-line `5h xx% / 7d yy%` legend stay vertically centered around the same y.
- [ ] Crash recovery still works: send the process SIGKILL (`kill -9`) → launchd respawns it within ~2s.

### DIFF
--- a/panel/Bootstrap.swift
+++ b/panel/Bootstrap.swift
@@ -673,9 +673,13 @@ enum Bootstrap {
         let python = venvURL.appendingPathComponent("bin/python3").path
         let stackvox = venvURL.appendingPathComponent("bin/stackvox").path
         let logPath = "\(installDir)/daemon.log"
+        // Daemon mirrors install.sh's "always" mode — if stackvox serve ever
+        // exits 0 launchd should still bring it back. Only the panel uses the
+        // SuccessfulExit:false form so a user Quit actually quits.
         try writePlist(label: daemonLabel,
                        programArgs: [python, stackvox, "serve"],
                        logPath: logPath,
+                       keepAlive: true,
                        env: stackvoxEnv(venvURL: venvURL))
     }
 
@@ -694,17 +698,20 @@ enum Bootstrap {
 
     // Common plist serialiser: emits the same XML shape install.sh's
     // register_launchd_agent function produces, via PropertyListSerialization.
+    //
+    // `keepAlive` mirrors install.sh's "always" vs "on_crash" modes:
+    //   true                       → restart unconditionally
+    //   ["SuccessfulExit": false]  → restart on crash only (Quit means quit)
     private static func writePlist(label: String,
                                    programArgs: [String],
                                    logPath: String,
+                                   keepAlive: Any = ["SuccessfulExit": false],
                                    env: [String: String] = [:]) throws {
         var plist: [String: Any] = [
             "Label":             label,
             "ProgramArguments":  programArgs,
             "RunAtLoad":         true,
-            // Restart only on crashes (non-zero exit). An explicit user
-            // Quit ends with a successful exit and should actually quit.
-            "KeepAlive":         ["SuccessfulExit": false],
+            "KeepAlive":         keepAlive,
             "StandardOutPath":   logPath,
             "StandardErrorPath": logPath,
         ]
--- a/panel/Bootstrap.swift
+++ b/panel/Bootstrap.swift
@@ -157,6 +157,27 @@ enum Bootstrap {
         // write them when the user finishes the wizard.
         retargetLaunchAgentIfNeeded(label: appLabel)
         retargetLaunchAgentIfNeeded(label: daemonLabel)
+        migrateKeepAliveIfNeeded(label: appLabel)
+    }
+
+    // Older installs wrote KeepAlive: true, so launchd respawned the panel
+    // even after an explicit user Quit. Rewrite to the dict form that
+    // restarts on crash only.
+    private static func migrateKeepAliveIfNeeded(label: String) {
+        let fm = FileManager.default
+        let plistPath = "\(launchAgentsDir)/\(label).plist"
+        guard fm.fileExists(atPath: plistPath),
+              let data = try? Data(contentsOf: URL(fileURLWithPath: plistPath)),
+              var plist = (try? PropertyListSerialization.propertyList(
+                  from: data, options: [], format: nil)) as? [String: Any]
+        else { return }
+        guard (plist["KeepAlive"] as? Bool) == true else { return }
+        plist["KeepAlive"] = ["SuccessfulExit": false]
+        guard let updated = try? PropertyListSerialization.data(
+            fromPropertyList: plist, format: .xml, options: 0) else { return }
+        try? updated.write(to: URL(fileURLWithPath: plistPath), options: [.atomic])
+        _ = try? runLaunchctl(["unload", plistPath])
+        _ = try? runLaunchctl(["load",   plistPath])
     }
 
     // Read the on-disk launchd plist for `label`; if its first program-
@@ -668,7 +689,9 @@ enum Bootstrap {
             "Label":             label,
             "ProgramArguments":  programArgs,
             "RunAtLoad":         true,
-            "KeepAlive":         true,
+            // Restart only on crashes (non-zero exit). An explicit user
+            // Quit ends with a successful exit and should actually quit.
+            "KeepAlive":         ["SuccessfulExit": false],
             "StandardOutPath":   logPath,
             "StandardErrorPath": logPath,
         ]
--- a/panel/Bootstrap.swift
+++ b/panel/Bootstrap.swift
@@ -78,6 +78,7 @@ enum Bootstrap {
     static let venvSymlinkPath  = "\(NSHomeDirectory())/.stack-nudge/venv"
     static let configPath       = "\(NSHomeDirectory())/.stack-nudge/config"
     static let phrasesDir       = "\(NSHomeDirectory())/.stack-nudge/phrases"
+    static let userQuitMarker   = "\(NSHomeDirectory())/.stack-nudge/user-quit"
 
     static let launchAgentsDir  = "\(NSHomeDirectory())/Library/LaunchAgents"
     static let appLabel         = "com.stackonehq.stack-nudge"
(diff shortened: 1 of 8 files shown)


panel/Bootstrap.swift
```diff
@@ -668,7 +689,9 @@ enum Bootstrap {
             "Label":             label,
             "ProgramArguments":  programArgs,
             "RunAtLoad":         true,
-            "KeepAlive":         true,
+            // Restart only on crashes (non-zero exit). An explicit user
+            // Quit ends with a successful exit and should actually quit.
+            "KeepAlive":         ["SuccessfulExit": false],
```
StuBehan: `writePlist` is shared by both `writePanelPlist` and `writeDaemonPlistIfVenvPresent`, so this hardcoded `SuccessfulExit:false` also lands on the **voice daemon** plist when a user installs via the in-app wizard (`install` → `writeDaemonPlistIfVenvPresent`). But `install.sh` intentionally keeps the daemon on `always` and only moves the app to `on_crash`. So the daemon's restart policy now diverges by install method, and if `stackvox serve` ever exits 0 launchd won't bring it back. The comment just above ("KeepAlive + RunAtLoad mirror install.sh") is also no longer true for the daemon.

Suggest parameterizing the mode like `register_launchd_agent` does:
```swift
private static func writePlist(label: String,
                               programArgs: [String],
                               logPath: String,
                               keepAlive: Any = ["SuccessfulExit": false],
                               env: [String: String] = [:]) throws {
```
and passing `keepAlive: true` from `writeDaemonPlistIfVenvPresent`.

panel/Bootstrap.swift
```diff
@@ -157,6 +157,27 @@ enum Bootstrap {
...
+              var plist = (try? PropertyListSerialization.propertyList(
+                  from: data, options: [], format: nil)) as? [String: Any]
+        else { return }
+        guard (plist["KeepAlive"] as? Bool) == true else { return }
+        plist["KeepAlive"] = ["SuccessfulExit": false]
+        guard let updated = try? PropertyListSerialization.data(
+            fromPropertyList: plist, format: .xml, options: 0) else { return }
+        try? updated.write(to: URL(fileURLWithPath: plistPath), options: [.atomic])
+        _ = try? runLaunchctl(["unload", plistPath])
+        _ = try? runLaunchctl(["load",   plistPath])
```
StuBehan, line 193: `unload` of the panel's own agent SIGTERMs this running process; the `load` + `RunAtLoad` then respawns it — a one-time self-restart on first launch after update. Same pattern as `retargetLaunchAgentIfNeeded`, but that only fired for the narrow pre-1.7 path population whereas this fires for every ≤1.20.0 user. Two things worth confirming on a real ≤1.20.0 install: (1) the unload→load window can't leave the panel down if SIGTERM lands before `load` re-registers, and (2) we don't briefly end up with an orphaned old process alongside the new `RunAtLoad` one (install.sh guards this with `pkill`; this path has no equivalent). The reload is genuinely needed for Quit to take effect without a re-login, so the trade-off is fine — just want it to be a deliberate call.

panel/SessionUsage.swift
```diff
@@ -99,6 +106,14 @@ final class QuotaProbe {
     }
 
     private func fetch(retried: Bool, completion: @escaping (QuotaSnapshot?) -> Void) {
+        if let until = retryAfterUntil, until > Date() {
+            // Still inside the 429 backoff window. Don't fire the request —
+            // it would re-trigger the rate limit and reset our cooldown.
+            lastProbeFailed = true
```
StuBehan: Setting `lastProbeFailed = true` here (and on the real 429 at line 160) makes `runQuotaProbe` show `"Quota data unavailable — the usage endpoint may have changed."` (`Panel.swift:1239`). During rate-limiting the endpoint hasn't changed, so that message is misleading — and our own standard is that error messages should accurately describe what happened. Consider a distinct rate-limited state/message (e.g. "Rate-limited — retrying shortly"). The backoff mechanics themselves look correct.

panel/SessionUsage.swift
```diff
@@ -138,19 +153,44 @@ final class QuotaProbe {
...
+                self?.retryAfterUntil = nil
                 completion(snapshot)
             }
         }.resume()
     }
 
+    // RFC 7231: Retry-After is either an HTTP-date or a non-negative integer
+    // delta-seconds. Anthropic typically returns the integer form; tolerate
+    // either, and clamp to a max so a misconfigured value can't strand us.
+    private static func parseRetryAfter(_ http: HTTPURLResponse?) -> TimeInterval? {
```
StuBehan: `parseRetryAfter` is pure and a natural unit-test target — delta-seconds, HTTP-date, negative value, the >1h clamp, and garbage→nil. Worth a small test even though `swift test` needs Xcode locally.

panel/CompactView.swift
```diff
@@ -115,8 +115,12 @@ struct CompactView: View {
                 )
                 .frame(width: size, height: size)
             }
-
+            // Force sideText to fill the gauge cluster's height so its
+            // ZStack centers at the same y as the gauge's center digit,
+            // regardless of which child (countdown vs two-line legend)
+            // is currently visible.
             sideText
+                .frame(height: size + 2)
```
StuBehan: `size + 2` is duplicated from the halo `Circle`'s frame (line 103). Since the whole point is that these two stay equal, consider a shared constant (or derive one from the other) so they can't drift apart later. The explicit `HStack(alignment: .center)` above is also redundant (`.center` is the default), though harmless as intent documentation.

panel/Bootstrap.swift
hiskudin: Good catch — fixed in 01dd78a. `writePlist` now takes a `keepAlive` parameter (defaults to `["SuccessfulExit": false]` for the panel), and `writeDaemonPlistIfVenvPresent` passes `keepAlive: true` so the daemon keeps install.sh's "always" semantics. Updated the comment block too.

panel/Bootstrap.swift
hiskudin: Worth thinking through. The two concerns:

**(1) SIGTERM landing before `load` re-registers.** `NSApplication` installs its own SIGTERM handler that dispatches through the main run loop (`applicationShouldTerminate` → `applicationWillTerminate`); it doesn't bypass the run loop and call `exit()` synchronously. `runLaunchctl` blocks the main thread on `Process.waitUntilExit()`, so the SIGTERM notification queues up but doesn't fire until we return from the migration. Both `unload` and `load` complete before AppKit gets to terminate us — at which point launchd has already brought up the new instance via `RunAtLoad`.

**(2) Orphaned old + new RunAtLoad.** Unlike `install.sh`'s reinstall path (which uses `pkill` to clear the old process), here the old process *is* us — `unload` SIGTERMs us deterministically, and there's only ever one panel at a time because we're a singleton agent tied to the bundle-id label. The brief overlap between `load` spawning a new instance and our SIGTERM landing is a few milliseconds at most; not visible to the user.

The migration runs once per user (we guard on `KeepAlive == true`), so even if a worst-case race did happen, the next login would still come up cleanly via the rewritten plist. I'd rather leave the explicit Quit-takes-effect-immediately UX as the priority and accept this trade-off.

panel/SessionUsage.swift
hiskudin: Fair point — fixed in 01dd78a. Added a separate `isRateLimited` getter on `QuotaProbe` that reads off `retryAfterUntil`; `lastProbeFailed` now only flags real failures (non-200, non-429). The Usage tab branches on `isRateLimited` first and shows "Rate-limited by Anthropic — retrying shortly." before falling through to the "endpoint may have changed" message. Both 429 paths (cooldown skip + live 429) now stop setting `lastProbeFailed = true`.

panel/SessionUsage.swift
hiskudin: Done in 01dd78a. Refactored `parseRetryAfter` to take `String?` so tests don't need an `HTTPURLResponse` (the `HTTPURLResponse?` overload is now a one-liner wrapper). New `QuotaProbeRetryAfterTests` covers: nil, empty/whitespace, delta-seconds, negative-rejected, >1h clamp via `maxRetryAfter`, HTTP-date in the future (±2s drift tolerance), HTTP-date in the past → 0, and garbage → nil.

panel/CompactView.swift
hiskudin: The original duplicate (the `VStack { Spacer; sideText; Spacer }.frame(height: size + 2)` from 0804c37) is gone — the layout rewrite in 8e73847 replaced sideText's centering with a measured-width slot that doesn't reference `size + 2`. The halo's `size + 2` is the only remaining occurrence, but 01dd78a still extracts it to a local `haloSize = size + 2` so a future addition can't drift, and drops the redundant explicit `HStack(alignment: .center)`.

StuBehan (approved):
LGTM

PR #111, opened 2026-06-30
fix(quota): read Claude quota via `claude /usage` instead of keychain

hiskudin:
## Summary

- Replaces the direct \`/api/oauth/usage\` HTTP probe (which reads the OAuth token from the macOS Keychain) with a shell-out to \`claude --print --output-format json "/usage"\`. The CLI is a client-side intercept (\`num_turns: 0\`, \`duration_api_ms: 0\`, \`total_cost_usd: 0\`) that hits the same backend but runs inside Claude Code, which has its own keychain ACL grant — so StackNudge's process never touches the keychain and the periodic 8h password prompt is gone.
- Keeps the legacy API probe as a fallback for missing-CLI / spawn-failure / parse-failure paths only. Rate-limited soft-fails hold the prior snapshot rather than re-triggering the keychain path.
- Pins the probe's cwd to \`~/.stack-nudge\` and unlinks the session rollout file (\`~/.claude/projects/<cwd>/<uuid>.jsonl\`) that every \`claude --print\` writes, so a 60s polling cadence doesn't litter \`claude --resume\` with ~1.4k empty sessions/day. UUID-shape guard on the unlink prevents any path-traversal escape.
- Settings → Usage now stays silent when the CLI probe is the source. The plaintext-file warning and the keychain-fallback explainer still surface in their respective modes (they have actionable info; the happy path doesn't).

Anthropic ToS (post Jan/Feb 2026) explicitly bans third-party OAuth tokens hitting \`/api/oauth/usage\`, so the CLI path is also the only policy-clean route long-term. Codex and Antigravity are unaffected (local rollout files and loopback RPC respectively — no keychain involved).

Closes the source of the recurring "stack-nudge wants to access Claude Code-credentials" prompt that anthropics/claude-code#22144 documents and Anthropic closed as not-planned.

## Test plan

- [x] \`./build.sh\` — clean compile under Developer ID signing.
- [x] Cold-start verification: killed all \`stack-nudge\` processes, swapped binary into \`~/Applications/StackNudge.app\`, relaunched fresh. Debug instrumentation (since removed) confirmed CLI probe fires first, finds \`/opt/homebrew/bin/claude\`, returns 1682 bytes parsed as \`.ok\` in ~4s. \`/usr/bin/log show --predicate 'processImagePath CONTAINS "stack-nudge"' --last 90s\` showed zero outbound traffic to \`api.anthropic.com\` — API probe correctly suppressed.
- [x] No password prompts during launch or first poll.
- [x] Session-file cleanup: after probe, \`~/.claude/projects/-Users-hiskud--stack-nudge/\` is empty (the directory itself disappears once Claude has nothing to keep there).
- [x] Encoding fix verified: Claude maps both \`/\` and \`.\` to \`-\` in the projects subdirectory name, so \`/Users/me/.stack-nudge\` becomes \`-Users-me--stack-nudge\` (note the double dash). \`probeSessionsDir\` matches.
- [ ] CI: 13 parser unit tests in \`ClaudeCliQuotaProbeTests.swift\` covering full-output, 0%-no-resets, malformed-line skip, unknown-bucket skip, envelope unwrapping, session-id extraction, path-traversal guard, and reset-time parsing with timezone. (Local \`swift test\` is blocked because this machine has Command Line Tools but not full Xcode — XCTest isn't on the toolchain.)

### DIFF
--- a/Tests/StackNudgePanelCoreTests/ClaudeCliQuotaProbeTests.swift
+++ b/Tests/StackNudgePanelCoreTests/ClaudeCliQuotaProbeTests.swift
@@ -1,0 +1,224 @@
+import XCTest
+
+@testable import StackNudgePanelCore
+
+final class ClaudeCliQuotaProbeTests: XCTestCase {
+
+    // MARK: - parseTierLine
+
+    func testParseTierLineWithResetsSuffix() {
+        let line = "Current session: 2% used · resets Jun 30 at 6:50pm (Europe/London)"
+        let parsed = ClaudeCliQuotaProbe.parseTierLine(line)
+        XCTAssertEqual(parsed?.name, "session")
+        XCTAssertEqual(parsed?.tier.utilization, 2)
+        XCTAssertNotNil(parsed?.tier.resetsAt)
+    }
+
+    func testParseTierLineWithoutResetsSuffix() {
+        // 0% lines omit the "· resets" suffix.
+        let line = "Current week (Sonnet only): 0% used"
+        let parsed = ClaudeCliQuotaProbe.parseTierLine(line)
+        XCTAssertEqual(parsed?.name, "week (Sonnet only)")
+        XCTAssertEqual(parsed?.tier.utilization, 0)
+        XCTAssertNil(parsed?.tier.resetsAt)
+    }
+
+    func testParseTierLineRejectsMalformed() {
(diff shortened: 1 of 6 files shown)


StuBehan (approved):
LGTM

PR #114, opened 2026-07-01
feat(panel): jump to top/bottom with cmd+up/down

StuBehan:
## Summary

Adds ⌘↑ / ⌘↓ to jump the current page to its first / last item, from any
scroll position — on every list page in the panel.

## Changes

- One `⌘↑/↓` case in the existing `cmdOnly` block of `panelHandlesKey`,
  dispatching per mode through a new `jumpToEdge(top:)`:
  - **Events / Sessions / Tickets / Settings / Phrases** move the selection to
    the first/last row — the viewport follows, since each page already scrolls
    to its selection.
  - **Usage** — the client list jumps to first/last client; the detail pane
    (no selection) scrolls the clip view to the top/bottom edge.
  - **Modal / single-purpose screens** (bootstrap, uninstall, update,
    post-update) return `false`, so the keystroke passes through untouched.
- `selectFirst`/`selectLast` helpers on `EventStore`, `Phrases`, and `PanelNav`
  (settings rows, outcomes index, usage client), plus a `scrollDetailToEdge`
  analog to the existing `scrollDetailBy`.
- Each page footer gains a **`Top/Bottom ⌘↑↓`** hint.

## Testing

- `./build.sh` → **Build complete**, zero warnings (rebased onto 1.20.3).
- `EventStoreTests` and `SettingsRowTests` cover the first/last jump helpers
  (incl. empty-store); run in CI via `swift test`.
- Needs a manual eyeball (the panel can't be driven in CI): ⌘↑/↓ on each tab —
  lands on the first/last row, and the Usage detail pane scrolls to its edges.

## Related issues

Companion to the Tickets-tab cursor-nav smoothing (#113). No tracked issue.

### DIFF
--- a/Tests/StackNudgePanelCoreTests/EventStoreTests.swift
+++ b/Tests/StackNudgePanelCoreTests/EventStoreTests.swift
@@ -26,6 +26,24 @@ final class EventStoreTests: XCTestCase {
         XCTAssertEqual(store.selectedID, event.id)
     }
 
+    func test_selectFirstAndLast_jumpToEnds() {
+        let store = EventStore()
+        let oldest = makeEvent(message: "oldest")
+        let newest = makeEvent(message: "newest")
+        store.append(oldest)
+        store.append(newest)   // newest-first: events == [newest, oldest]
+        store.selectLast()
+        XCTAssertEqual(store.selectedID, oldest.id)   // bottom row
+        store.selectFirst()
+        XCTAssertEqual(store.selectedID, newest.id)   // top row
+    }
+
+    func test_selectFirst_onEmptyStore_clearsSelection() {
+        let store = EventStore()
+        store.selectFirst()
+        XCTAssertNil(store.selectedID)
+    }
+
     func test_append_truncatesPastMaxEvents() {
         // maxEvents is 5; we send 7.
         let store = EventStore()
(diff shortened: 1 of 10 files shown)


hiskudin (commented):
Reviewed the diff. Small, focused, tested — no changes needed from me.

- Dispatch table in `jumpToEdge` is the right shape: `.sessions` guards `renamingPID == nil` before intercepting (so ⌘↑/↓ passes through into the rename field where it's expected as text-nav), and modal/single-purpose screens fall through the `default` returning `false`.
- Usage split (`usageDetailFocused` → `scrollDetailToEdge`, otherwise client list first/last) mirrors the existing ↑↓ handling. `scrollDetailToEdge` clamps `maxY = max(0, doc.frame.height - clip.bounds.height)` correctly for the short-content case.
- Empty-store test on `EventStore.selectFirst` is a nice touch — asserts `selectedID == nil` rather than crashing.

Tiny style nit only (not worth a change): `selectFirstUsageClient` / `selectLastUsageClient` are semicolon-chained one-liners while `selectFirstRow` / `selectLastRow` a few functions later use guarded assignment across lines. Trivially inconsistent.

Ship it.

hiskudin (approved):

PR #130, opened 2026-07-14
feat(panel): timed global mute + CLI-only Claude usage probe

hiskudin:
## Summary

This branch carries two independent changes.

### 1. Timed global mute for notifications (`feat`)

A "mute everything for X minutes" control that suppresses **every** interrupting output — banner, sound, voice, and the activate-immediately focus jump — for all events **including permission prompts** — then auto-lifts when the timer expires. Events still flow into the panel while muted.

- Transient: in-memory only, resets on relaunch. A single gate at the top of `postBannerIfNeeded` does the suppression.
- The default one-click duration is a config-backed preference (`STACKNUDGE_MUTE_DURATION_MIN`, allowed 15/30/60/120, default 30).

**Surfaces**
- Panel header **bell button** with live countdown
- Menu-bar **submenu** (For 15m / 30m / 1h / 2h) + `bell.slash` icon + countdown badge while muted
- **Settings** → Toggles: a Mute/Resume action row (bell glyph, not a nav chevron), a muted-only top banner with one-click Resume, and the duration cycle row
- **Compact widget** `bell.slash` glyph while muted
- **`M` keyboard shortcut** in the Events tab, with footer hint

### 2. Drop keychain/API Claude usage probe, rely solely on the `claude` CLI (`refactor`)

Removes `QuotaProbe` — the only code that read the Claude OAuth token from the **macOS Keychain** (`SecItemCopyMatching`) or the **plaintext `~/.claude/.credentials.json`**, then called the **unofficial `api.anthropic.com/api/oauth/usage`** endpoint directly.

Claude quota now comes **solely** from `claude --print /usage` (`ClaudeCliQuotaProbe`), which reads its *own* keychain — so StackNudge never touches the keychain or hits the API, eliminating the periodic password prompts caused by Claude Code rotating its keychain item.

Also removes the now-dead `usingPlaintextCredentials` / `usingClaudeCliProbe` flags, the Settings "probe source" status label, and `QuotaProbeRetryAfterTests`; rewords the empty-state copy and comments (no more "unofficial endpoint" / keychain-prompt language).

> [!WARNING]
> **User-visible behavioral change:** without the `claude` CLI on PATH / signed in, the Usage tab now shows *"Claude usage unavailable — run `claude /usage` to check your session"* instead of silently falling back to the keychain/API. Fine for this app's audience (it exists to watch Claude Code), but a conscious trade-off. Codex/Antigravity probes and GitHub's keychain use are **untouched**.

## Docs

`README.md` brought in line with both changes:

- **Mute** — new "Mute for a while" note plus the `M` row in the Events-tab table, the `Mute notifications ▸` menu-bar row, the Settings-tab Mute/Resume + duration row, and the compact-widget glyph (no countdown there — corrected).
- **Quota source** — rewrote the Usage-tab data-source paragraph (no more `/api/oauth/usage` / keychain "Always Allow" copy) and softened the uninstall note now that we create no keychain entries.
- **Drive-by fix (pre-existing):** the tab list was wrong — the panel has **five** tabs (Tickets was missing), Tickets is `Cmd+4`, and **Settings is `Cmd+5`** (was documented as `Cmd+4`).

## Testing

- `./build.sh` clean + Developer-ID signed after each change and after merging `origin/main`.
- Added `MuteStateTests` (transient state, mute-row order, countdown-label rounding); removed the obsolete `QuotaProbeRetryAfterTests`; `ClaudeCliQuotaProbeTests` unchanged.
- Mute feature installed and exercised locally: bell, `M` key, menu submenu, Settings row + banner, and confirmed actual suppression of banner/sound/voice/permission-prompt while events still land in the panel and the timer auto-lifts.


panel/Panel.swift
```diff
@@ -1243,48 +1281,25 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
-                } else if self.quotaProbe.isRateLimited {
-                    self.nav.quotaError = "Rate-limited by Anthropic — retrying shortly."
-                } else if self.quotaProbe.lastProbeFailed {
-                    self.nav.quotaError = "Quota data unavailable — the usage endpoint may have changed."
-                }
+            } else if self.claudeCliQuotaProbe.isRateLimited {
+                // Soft-fail: leave the prior snapshot in place, no error.
+            } else {
+                // Hard-fail: the CLI couldn't run or its output didn't parse.
+                self.nav.quotaError = "Claude usage unavailable — run `claude /usage` to check your session."
```
StuBehan, line 1312: Hard-fail sets `quotaError` but leaves `nav.quota` populated. `emptyState` (the error surface) only renders when `availableUsageClients` is empty, and a stale Claude snapshot keeps `.claude` available (`PanelNav.swift:469`), so the old bars keep rendering and this error is never shown. After the `claude` CLI is removed/signed out, the tab shows stale usage as current with no error.

Fix: clear `nav.quota` and `quotaLastUpdated` on hard-fail, or thread a stale/error flag into the tier view.

panel/Panel.swift
```diff
@@ -1243,48 +1281,25 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
-                    self.nav.quota = snapshot
-                    self.nav.quotaError = nil
-                    self.nav.quotaLastUpdated = Date()
-                    self.evaluateQuotaThresholds(snapshot)
-                } else if self.quotaProbe.isRateLimited {
-                    self.nav.quotaError = "Rate-limited by Anthropic — retrying shortly."
-                } else if self.quotaProbe.lastProbeFailed {
-                    self.nav.quotaError = "Quota data unavailable — the usage endpoint may have changed."
-                }
+            } else if self.claudeCliQuotaProbe.isRateLimited {
```
StuBehan, line 1298: First probe with no prior snapshot: a rate-limited soft-fail leaves `quota` and `quotaError` both nil. With no Codex/Antigravity client, the tab shows "Loading usage…" until a non-limited tick, with no rate-limit indication. Set a transient rate-limited state on cold soft-fail.

panel/Panel.swift
```diff
@@ -2007,6 +2029,41 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+
+    // Fires every 30s while muted: refreshes the countdown UI and lifts the
+    // mute once the window passes. Nil when not muted.
+    private var muteTicker: Timer?
+
+    // Suppress all banner/sound/voice output for `minutes`, then auto-lift.
+    // The state (nav.muteUntil) is transient, so a relaunch clears it. Muting
+    // also dismisses any banner already on screen, so it silences an in-flight
+    // nudge rather than only future ones.
+    func muteFor(minutes: Int) {
```
StuBehan, line 2052: `muteFor` clears delivered banners but doesn't stop an in-flight voice utterance — a `Speaker.speak(...)` already playing continues to completion. Add `Speaker.stopAllAudio()` (as on quit, :2307).

panel/PanelNav.swift
```diff
@@ -182,6 +186,27 @@ final class PanelNav: ObservableObject {
...
     @Published var muteWhenFocused: Bool = true
+    // Timed global mute. When `muteUntil` is a future date, PanelController
+    // suppresses ALL banner/sound/voice output (permission prompts included)
+    // until it passes, then auto-lifts. Deliberately transient — never read
+    // in loadFromConfig and never written to disk, so it resets on relaunch.
+    // `muteTick` is bumped by PanelController's 30s countdown timer purely to
+    // re-render the tab-strip/pill countdown while muted (the SwiftUI views
+    // reference it so a value-less change still refreshes them).
+    @Published var muteUntil: Date?
+    @Published var muteTick: Int = 0
```
StuBehan, line 197: Non-blocking cleanup:

- `muteTick` + the 30s `Timer` re-implement the existing `TimelineView(.periodic…)` pattern. `isMuted`/`muteRemainingLabel` are pure functions of `muteUntil`; a `TimelineView` derives both and drops ~30 lines plus the three `let _ = nav.muteTick` lines (deleting one silently freezes that countdown, no warning).
- Duration label rendered 3 ways: "1h"/"2h" (menu), "60 min" (Settings), "60m" (bell/badge). Use one shared `muteDurationLabel(minutes:)`.
- Muted-bell visual hand-built 4× (pill/header/Settings/menu-bar); `CompactView` uses `.secondary`, others use `.accentColor`. Extract one `MutedBell` view.
- `QuotaProbeRetryAfterTests` deleted; new probe hardcodes a 60s backoff with no test.
- Nearest-option snapping duplicated with a dead `?? 30` fallback; cycle-advance index math copy-pasted ~9×; `refreshMuteBadge` rebuilds the `NSImage` every tick when only the title changes.

StuBehan (commented):
Two correctness gaps, both small fixes; plus non-blocking cleanup, detailed inline.

**1. `Panel.swift:2081` — snooze re-fire bypasses the mute gate.** `snoozeEvent`'s timer re-posts via `postBanner(for:)`, skipping the `if nav.isMuted { return }` gate in `postBannerIfNeeded`. A snoozed prompt fires a banner during an active mute window — the only involuntary banner path the gate doesn't cover. Fix: re-post through `postBannerIfNeeded(current.with(snoozedUntil: nil))` (also picks up session-mute/focus handling), or guard with `guard !nav.isMuted else { return }`.

**2. `Panel.swift:1302`** — hard-fail quota error is never shown when a stale snapshot exists (inline).

Note: the global-mute gate also suppresses `bypassMute` (welcome/critical) events. The comment documents this as intended — flagging to confirm it's deliberate.

panel/Panel.swift
hiskudin: Fixed in 400e60d — the hard-fail branch now clears `nav.quota` + `nav.quotaLastUpdated`, so once the `claude` CLI is removed/signed out the Usage tab surfaces the error state instead of rendering the stale snapshot as current. Good catch; this became reachable exactly because the API fallback that previously masked it is gone.

panel/Panel.swift
hiskudin: Fixed in 400e60d — on a cold rate-limited soft-fail (`nav.quota == nil`) we now set a transient `quotaError` ("Claude usage rate-limited — retrying shortly") so the tab shows that rather than an indefinite spinner. Warm soft-fails still hold the prior snapshot untouched, no error.

panel/Panel.swift
hiskudin: Fixed in 400e60d — `muteFor` now calls `Speaker.stopAllAudio()` alongside `removeAllDeliveredNotifications()`, so an in-flight utterance is cut immediately. Also corrected the comment, which claimed to silence in-flight nudges but only dropped the banner.

panel/PanelNav.swift
hiskudin: Thanks — all fair. Capturing these as follow-ups (none block this PR):

- Shared `muteDurationLabel(minutes:)` for the 3 formats.
- Extract a single `MutedBell` view and reconcile CompactView's `.secondary` vs the others' `.accentColor`.
- A test for the CLI probe's hardcoded 60s backoff.
- Drop the dead `?? 30`; cache the `refreshMuteBadge` NSImage (rebuild only when the title changes).

On `TimelineView` vs `muteTick`: only partly applicable — the controller timer also drives the menu-bar `NSStatusItem` badge and the auto-lift side effect, neither of which a `TimelineView` can do, so it can't fully replace the ticker (it could still drop the in-view `let _ = nav.muteTick` reads). The cycle-index math duplication predates this PR.

I'll land these in a separate cleanup PR unless you'd rather see them here.

hiskudin:
Thanks for the thorough review @StuBehan. Pushed `400e60d` covering the correctness items (inline replies on each thread):

- **Snooze re-fire bypassing the mute gate** — fixed. The snooze timer clears the snooze flag, then `guard !nav.isMuted else { return }` before re-posting, so a snoozed nudge can't fire a banner during a mute window. I kept it banner-only rather than routing through `postBannerIfNeeded(...)` — that would turn a re-fire into a full re-notification (sound/voice/activate-immediately), a behavior change beyond the fix.
- **Stale-snapshot hard-fail**, **cold rate-limit spinner**, **in-flight voice** — all fixed (see inline).

On the **`bypassMute` note**: yes, suppressing welcome/critical events during an active mute is intentional. The spec for this feature is "mute *everything*, permission prompts included," for a bounded window that auto-lifts — a user who explicitly muted for N minutes shouldn't be pierced by a welcome banner. Happy to carve out a truly-critical class later if we decide one warrants it, but that'd be a deliberate follow-up, not part of this window's contract.

Cleanup suggestions captured as follow-ups (noted inline on `PanelNav.swift`).

StuBehan (approved):
LGTM

PR #159, opened 2026-08-20
feat(widget): free placement for the compact widget

faisalreza-stackone:
## What

Adds an opt-in "free placement" mode for the compact widget (the always-on pill), which until now could only sit at one of four screen corners.

A new **Snap to corners** toggle in Settings → Widget drives it. It defaults to **On**, so existing users see no change. Turn it off and the pill stays wherever you drop it.

## Behaviour

- **Snap on (default):** unchanged. Releasing a drag animates the pill to the nearest corner and persists that corner.
- **Snap off:** the pill stays where you drop it. Its position is saved and restored across relaunches.
- **Own display:** a free-placed pill stays on the screen you left it on, not whichever screen the cursor happens to be on.
- **Dock level:** the pill clamps into the screen's full frame rather than the Dock-excluded area, so it can rest down at Dock level. The top stays capped just below the menu bar, and nothing hangs off the physical screen.
- **Re-enabling snap** sends the pill to the corner nearest its last free spot. **Toggling snap off** leaves it where it currently rests.

## Settings

New **Snap to corners** toggle. The existing **Widget corner** row dims while snap is off, since it has no effect in free mode.

## Config

- `STACKNUDGE_COMPACT_SNAP` (`true`/`false`, default `true`)
- `STACKNUDGE_COMPACT_POS` (`x,y`, the saved free origin)

`STACKNUDGE_COMPACT_CORNER` is unchanged and remains the source of truth while snap is on.

## Scope

The pill widget only. Notification banners are macOS notifications positioned by the system, and the full expanded panel keeps its own position handling. Neither is touched.

## Testing

- Unit tests for the new geometry and persistence helpers (`CompactPlacement`): nearest-corner selection, clamping, screen selection, placement bounds, and position parse/format.
- Full `swift test` suite passes.
- Built and exercised in the live app: drag to free-place, Dock-level placement, snap on/off transitions, and multi-monitor placement.

### DIFF
--- a/Tests/StackNudgePanelCoreTests/CompactPlacementTests.swift
+++ b/Tests/StackNudgePanelCoreTests/CompactPlacementTests.swift
@@ -1,0 +1,80 @@
+import XCTest
+import CoreGraphics
+
+@testable import StackNudgePanelCore
+
+final class CompactPlacementTests: XCTestCase {
+
+    // A 1000x1000 frame at origin (0,0); midpoint is (500, 500).
+    private let frame = CGRect(x: 0, y: 0, width: 1000, height: 1000)
+
+    // MARK: - nearestCorner (AppKit coords: y grows upward)
+
+    func test_nearestCorner_quadrants() {
+        // Below-mid Y = bottom; left/right by X.
+        XCTAssertEqual(CompactPlacement.nearestCorner(toCenter: CGPoint(x: 100, y: 100), in: frame), .bottomLeft)
+        XCTAssertEqual(CompactPlacement.nearestCorner(toCenter: CGPoint(x: 900, y: 100), in: frame), .bottomRight)
+        XCTAssertEqual(CompactPlacement.nearestCorner(toCenter: CGPoint(x: 100, y: 900), in: frame), .topLeft)
+        XCTAssertEqual(CompactPlacement.nearestCorner(toCenter: CGPoint(x: 900, y: 900), in: frame), .topRight)
+    }
+
+    func test_nearestCorner_respectsFrameOrigin() {
+        // Frame shifted to a second display; midpoint is (2500, 500).
+        let shifted = CGRect(x: 2000, y: 0, width: 1000, height: 1000)
+        XCTAssertEqual(CompactPlacement.nearestCorner(toCenter: CGPoint(x: 2100, y: 900), in: shifted), .topLeft)
+        XCTAssertEqual(CompactPlacement.nearestCorner(toCenter: CGPoint(x: 2900, y: 100), in: shifted), .bottomRight)
+    }
+
+    // MARK: - clamp
+
+    func test_clamp_insideFrame_isUnchanged() {
+        let p = CompactPlacement.clamp(origin: CGPoint(x: 400, y: 400),
+                                       size: CGSize(width: 100, height: 50),
+                                       into: frame, inset: 14)
+        XCTAssertEqual(p, CGPoint(x: 400, y: 400))
+    }
+
+    func test_clamp_pullsBackWithinInsetOnAllSides() {
+        let size = CGSize(width: 100, height: 50)
+        // Off the bottom-left.
+        XCTAssertEqual(CompactPlacement.clamp(origin: CGPoint(x: -50, y: -50),
+                                              size: size, into: frame, inset: 14),
+                       CGPoint(x: 14, y: 14))
+        // Off the top-right (maxX = 1000-100-14 = 886, maxY = 1000-50-14 = 936).
+        XCTAssertEqual(CompactPlacement.clamp(origin: CGPoint(x: 5000, y: 5000),
+                                              size: size, into: frame, inset: 14),
+                       CGPoint(x: 886, y: 936))
+    }
+
+    func test_clamp_degenerateFrameSmallerThanPill_pinsToMinInset() {
+        // Frame narrower/shorter than pill+insets: pin to min inset, no NaN/flip.
+        let tiny = CGRect(x: 0, y: 0, width: 80, height: 40)
+        let p = CompactPlacement.clamp(origin: CGPoint(x: 999, y: 999),
+                                       size: CGSize(width: 100, height: 50),
+                                       into: tiny, inset: 14)
+        XCTAssertEqual(p, CGPoint(x: 14, y: 14))
+    }
+
+    // MARK: - parse / format
+
+    func test_parsePosition_roundTrip() {
+        let p = CGPoint(x: 123, y: 456)
+        XCTAssertEqual(CompactPlacement.parsePosition(CompactPlacement.formatPosition(p)), p)
+    }
+
+    func test_parsePosition_toleratesWhitespace() {
+        XCTAssertEqual(CompactPlacement.parsePosition(" 12 , 34 "), CGPoint(x: 12, y: 34))
+    }
+
+    func test_parsePosition_rejectsMalformedAndNil() {
+        XCTAssertNil(CompactPlacement.parsePosition(nil))
+        XCTAssertNil(CompactPlacement.parsePosition(""))
+        XCTAssertNil(CompactPlacement.parsePosition("123"))
+        XCTAssertNil(CompactPlacement.parsePosition("a,b"))
+        XCTAssertNil(CompactPlacement.parsePosition("1,2,3"))
+    }
+
+    func test_formatPosition_roundsToIntegers() {
+        XCTAssertEqual(CompactPlacement.formatPosition(CGPoint(x: 12.6, y: 34.2)), "13,34")
+    }
+}
--- a/Tests/StackNudgePanelCoreTests/CompactPlacementTests.swift
+++ b/Tests/StackNudgePanelCoreTests/CompactPlacementTests.swift
@@ -99,4 +99,17 @@ final class CompactPlacementTests: XCTestCase {
     func test_formatPosition_roundsToIntegers() {
         XCTAssertEqual(CompactPlacement.formatPosition(CGPoint(x: 12.6, y: 34.2)), "13,34")
     }
+
+    // MARK: - placementBounds (free placement: Dock free at bottom, menu bar capped at top)
+
+    func test_placementBounds_bottomIsPhysical_topIsBelowMenuBar() {
+        // Full screen 1000 tall; 60pt Dock at the bottom, 25pt menu bar at the top.
+        let physical = CGRect(x: 0, y: 0, width: 1000, height: 1000)
+        let visible  = CGRect(x: 0, y: 60, width: 1000, height: 915) // maxY = 975
+        let b = CompactPlacement.placementBounds(frame: physical, visibleFrame: visible)
+        XCTAssertEqual(b.minY, 0)      // reaches the physical bottom (over the Dock)
+        XCTAssertEqual(b.maxY, 975)    // capped at the menu-bar underside
+        XCTAssertEqual(b.minX, 0)      // full width
+        XCTAssertEqual(b.maxX, 1000)
+    }
 }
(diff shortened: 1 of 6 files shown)


panel/Panel.swift
```diff
@@ -1181,10 +1186,94 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+            let corner = CompactPlacement.nearestCorner(toCenter: center, in: visible)
+            if nav.compactCorner != corner {
+                nav.compactCorner = corner
+                ConfigFile.write(key: "STACKNUDGE_COMPACT_CORNER",
+                                 value: corner.rawValue)
+            }
+        } else {
+            nav.setCompactFreeOrigin(cornerOrigin(for: size, corner: nav.compactCorner))
+        }
+        applyCompactLayout()
```
hiskudin: **(1) This reads a stale `compactSnap`.**

`@Published` publishes in `willSet`, so while this sink is running `nav.compactSnap` still holds the **old** value. I verified it with a standalone Combine harness against the exact `.removeDuplicates().dropFirst().sink` shape used at `Panel.swift:840`:

```
sink received: false; reading nav.snap gives: true
```

That means the ternary at `Panel.swift:1045` picks `compactCornerOrigin` when the user just turned snap *off*, and `resolvedFreeOrigin` when they turned it *on* — inverted.

It's masked today: this toggle is only reachable from Settings, where `compactExpanded == true`, so `applyCompactLayout` takes the full-panel branch and never evaluates that ternary. The next collapse re-runs it with the committed value and the end state is correct. But it's one new entry point away from being live — a menu-bar item, a hotkey, or a config watcher that flips snap while the pill is collapsed would hit it immediately.

Worth closing now since it's cheap: thread `snap` down into the layout call the same way you did for `handleSnapModeChange`, or `DispatchQueue.main.async { self.applyCompactLayout() }` so the property has settled. The hazard is already documented in this file for `compactCorner` at `Panel.swift:1163`, which is what put me onto it.

panel/CompactPlacement.swift
```diff
@@ -0,0 +1,71 @@
...
+        CGRect(x: frame.minX, y: frame.minY,
+               width: frame.width, height: visibleFrame.maxY - frame.minY)
+    }
+
+    // Parse a persisted "x,y" string into a point. Returns nil for missing or
+    // malformed input so callers can fall back to a corner origin.
+    static func parsePosition(_ raw: String?) -> CGPoint? {
+        guard let raw else { return nil }
+        let parts = raw.split(separator: ",", omittingEmptySubsequences: false)
+        guard parts.count == 2,
```
hiskudin, line 61: **(2) `Double(_:)` accepts non-finite values.**

`Double("nan")`, `Double("inf")`, and `Double("-inf")` all parse successfully here. A hand-edited `STACKNUDGE_COMPACT_POS=nan,0` then propagates NaN straight through `clamp` — Swift's `min`/`max` pass NaN through rather than rejecting it, because every comparison against NaN is false — and into `panel.setFrame`.

The config file is user-editable by design (there's an *Open config file* row in Settings), so this is reachable without anyone doing anything exotic. One line:

```swift
guard parts.count == 2,
      let x = Double(parts[0].trimmingCharacters(in: .whitespaces)), x.isFinite,
      let y = Double(parts[1].trimmingCharacters(in: .whitespaces)), y.isFinite
else { return nil }
```

Worth a case in `test_parsePosition_rejectsMalformedAndNil` alongside the existing ones.

panel/Panel.swift
```diff
@@ -1181,10 +1186,94 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+            let center = NSPoint(x: origin.x + size.width / 2,
+                                 y: origin.y + size.height / 2)
+            let corner = CompactPlacement.nearestCorner(toCenter: center, in: visible)
+            if nav.compactCorner != corner {
+                nav.compactCorner = corner
+                ConfigFile.write(key: "STACKNUDGE_COMPACT_CORNER",
+                                 value: corner.rawValue)
+            }
+        } else {
+            nav.setCompactFreeOrigin(cornerOrigin(for: size, corner: nav.compactCorner))
```
hiskudin: **(3) Snap-off pins the free origin to the *cursor's* screen, not the pill's.**

`cornerOrigin` resolves against `activeScreen().visibleFrame` (`Panel.swift:1278`), which is the display under the mouse. Settings rows are keyboard-driven, so on a multi-monitor setup the cursor can easily be sitting on a different display than the panel when this fires — and the pill then lands on the wrong screen at the next collapse.

That contradicts the PR description's *"Toggling snap off leaves it where it currently rests"*, and it's the one case where the "own display" guarantee the rest of the change establishes doesn't hold.

`screenForFreeOrigin` — which you already added just below — is the fix; `cornerOrigin` just needs to take a screen instead of calling `activeScreen()` itself.

panel/Panel.swift
```diff
@@ -1033,7 +1042,9 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
             panel.contentMinSize = size
             var frame = panel.frame
             frame.size = size
-            frame.origin = compactCornerOrigin(for: size)
+            frame.origin = nav.compactSnap
```
hiskudin: **(4) Nit, free-mode placement is anchored bottom-left.**

`resolvedFreeOrigin` returns an origin, which in AppKit is the pill's bottom-left. Switching Widget type changes the size (290×56 for sessions, 145×66 for usage), so the left edge stays pinned and the pill shifts — whereas in snap mode with a right-hand corner the *right* edge stays put.

Practical effect: a free pill parked against the right edge leaves a ~145pt gap after switching to the usage widget. Cosmetic, and arguably fine to leave, but if you want parity you'd re-anchor on the nearer horizontal edge when the size changes.

panel/CompactPlacement.swift
```diff
@@ -0,0 +1,71 @@
...
+// isolation; PanelController supplies the concrete screen frame and window
+// size at call time.
+enum CompactPlacement {
+
+    // Which corner a pill centred at `center` is closest to, within
+    // `visibleFrame`. AppKit coordinates: y grows upward, so a center below
+    // the vertical midpoint is a *bottom* corner.
+    static func nearestCorner(toCenter center: CGPoint,
+                              in visibleFrame: CGRect) -> CompactCorner {
+        switch (center.x < visibleFrame.midX, center.y < visibleFrame.midY) {
```
hiskudin, line 16: Nit: the tie-break at exactly `midX`/`midY` is deterministic (falls to right/top, since `<` is false) but neither documented in the comment above nor covered by `test_nearestCorner_quadrants`. Worth one line either way — it's the kind of thing a later refactor flips silently.

Tests/StackNudgePanelCoreTests/CompactPlacementTests.swift
```diff
@@ -0,0 +1,115 @@
...
+        XCTAssertNil(CompactPlacement.parsePosition("1,2,3"))
+    }
+
+    func test_formatPosition_roundsToIntegers() {
+        XCTAssertEqual(CompactPlacement.formatPosition(CGPoint(x: 12.6, y: 34.2)), "13,34")
+    }
+
+    // MARK: - placementBounds (free placement: Dock free at bottom, menu bar capped at top)
+
+    func test_placementBounds_bottomIsPhysical_topIsBelowMenuBar() {
```
hiskudin, line 112: Nit: only exercised with a frame at origin `(0,0)`. The arithmetic (`visibleFrame.maxY - frame.minY`) is correct for a display arranged *below* the main one, where `frame.minY < 0` — but that's the case most likely to be got wrong later, and it's the one uncovered. `test_nearestCorner_respectsFrameOrigin` already sets the precedent for a shifted frame.

panel/Settings.swift
```diff
@@ -80,8 +80,9 @@ struct SettingsView: View {
                         }
 
                         section("Widget") {
-                            row(.widget,        label: "Widget",        kind: .toggle, value: nav.compactMode ? "On" : "Off")
-                            row(.widgetCorner,  label: "Widget corner", kind: .cycle,  value: nav.compactCorner.label,            enabled: nav.compactMode)
+                            row(.widget,        label: "Widget",         kind: .toggle, value: nav.compactMode ? "On" : "Off")
+                            row(.snapToCorners, label: "Snap to corners", kind: .toggle, value: nav.compactSnap ? "On" : "Off", enabled: nav.compactMode)
+                            row(.widgetCorner,  label: "Widget corner",   kind: .cycle,  value: nav.compactCorner.label,          enabled: nav.compactMode && nav.compactSnap)
                             row(.widgetOpacity, label: "Widget opacity", kind: .cycle, value: "\(Int(nav.compactAlpha * 100))%",  enabled: nav.compactMode)
```
hiskudin: Nit: the longer `"Snap to corners"` label re-aligned this row and the two above it, but `.widgetOpacity` below kept the old column spacing so the block no longer lines up.

hiskudin (commented):
Read the full diff in context, checked out the branch, and built it locally (`make build` clean; CI green across all six checks).

**Overall: solid.** Extracting the geometry into a pure `CompactPlacement` enum is the right call, and the tests are genuine — degenerate-frame clamp, malformed parse, and frame-origin offset are all covered. Default-on preserves existing behaviour. `.snapToCorners` is correctly wired into `settingsRows`, `applyCycle`, and `selectedRowRespondsToArrows`, so no half-wired row. Nice catch passing `snap` into `handleSnapModeChange` as a parameter rather than re-reading it — see the first inline comment for why that mattered.

Four things worth addressing; the two marked **(1)** and **(2)** are the ones I'd want fixed before merge. Details inline.

### Also: README not updated

Not anchorable to the diff, so noting it here. `README.md:210` still describes the widget settings as *"widget (corner, mascot picker, opacity)"*, and neither `STACKNUDGE_COMPACT_SNAP` nor `STACKNUDGE_COMPACT_POS` appears anywhere in the docs — the config table around `README.md:248` is where the other user-facing keys live. Repo convention is to keep it current; the last panel PR updated `README.md:189` for a smaller behaviour change than this one.

hiskudin (approved):

panel/Panel.swift
faisalreza-stackone: Fixed in 106bc50. `applyCompactLayout` now takes an optional `snap` override; the `$compactSnap` sink threads the new value through `handleSnapModeChange`, so layout no longer reads the stale `willSet` property. Kept it synchronous rather than dispatching async.

panel/CompactPlacement.swift
faisalreza-stackone: Fixed in 106bc50. `parsePosition` now requires `x.isFinite`/`y.isFinite`, so `nan`/`inf` are rejected before reaching `clamp`. Added `nan,0` / `0,inf` / `-inf,10` cases to `test_parsePosition_rejectsMalformedAndNil`.

panel/Panel.swift
faisalreza-stackone: Fixed in 106bc50. `cornerOrigin` now takes an optional `screen`; the snap-off branch seeds from `panel.screen` instead of `activeScreen()`, so keyboard toggling on multi-monitor keeps the widget on its own display.

panel/CompactPlacement.swift
faisalreza-stackone: Fixed in 106bc50. Documented the midX/midY tie-break (ties to right/top) in the comment and added an exact-midpoint assertion to `test_nearestCorner_quadrants`.

Tests/StackNudgePanelCoreTests/CompactPlacementTests.swift
faisalreza-stackone: Fixed in 106bc50. Added `test_placementBounds_displayBelowMain_negativeOrigin`, covering a frame with negative `minY` (a display arranged below the main one).

panel/Settings.swift
faisalreza-stackone: Fixed in 106bc50. Re-aligned the Widget rows, including `.widgetOpacity` and below.

panel/Panel.swift
faisalreza-stackone: Leaving this as-is. It's a cosmetic shift that only happens when switching Widget type while free-placed, and re-anchoring on the nearer edge adds size-change edge-detection for little practical gain. Noting it as a known tradeoff.

faisalreza-stackone:
On the README note: intentionally leaving the docs unchanged in this PR (author's call). Flagging so it's a conscious decision rather than an oversight — the widget-settings line at `README.md:210` and the two new keys (`STACKNUDGE_COMPACT_SNAP`, `STACKNUDGE_COMPACT_POS`) can follow in a separate docs pass if preferred.

Issue #163, opened 2026-09-04
Gemini CLI usage counter needs a Code Assist API call


hiskudin:
Split out of the compact-widget multi-client work (`feat/fixes-and-improvements`).

## Why it was left out

The widget and the Usage tab now cover Claude, Codex, and Antigravity. Gemini CLI is the one wired agent with no usage counter, and it is not an oversight — there is nothing local to read:

| Client | Source | Kind |
|---|---|---|
| Claude | `claude --print /usage` | shells out to the CLI, which reads its own keychain grant |
| Codex | newest `~/.codex/sessions/**/rollout-*.jsonl` → `payload.rate_limits` | local file |
| Antigravity | `agy` loopback Connect-RPC `GetUserStatus` | localhost, no auth |
| **Gemini** | **—** | **nothing** |

`~/.gemini` holds only `oauth_creds.json`, `settings.json`, `projects.json`, `state.json`, `google_accounts.json`, `installation_id`, `trustedFolders.json` and a `history/` dir. No rollouts with rate limits, no quota file, and the CLI serves no local endpoint the way `agy` does.

## What it would take

A live call to Google's Code Assist API using the OAuth token in `~/.gemini/oauth_creds.json`, including refresh handling.

That is a materially different shape of work from the other three probes and deserves its own review, because it would be **the first outbound network call this app makes to a non-local, non-Anthropic endpoint**. The Usage section of the README currently leans on the opposite property — that stack-nudge never touches your keychain and never calls a provider API directly, because the `claude` CLI does that for us and the other two probes are local reads. Reading Gemini quota over the network breaks that invariant, so the decision is a product one before it is an implementation one.

## Open questions

- Is the Code Assist quota endpoint documented, or would we be pinning an undocumented one that can change without notice?
- Token refresh: do we refresh in-process, or shell out to `gemini` and let it own its credentials (closer to how the Claude probe works, and keeps us out of the credential business)?
- Free-tier vs paid Code Assist — do both report a quota we can render on the same two-ring gauge?
- Does it need an opt-in setting, given it is the only probe that leaves the machine?

## Not blocking

`UsageClient` is exhaustive over its cases and the widget reads through `WidgetQuota.make`, so adding a `.gemini` case is a compile-error-guided change once a source exists.

PR #164, opened 2026-09-04
feat: multi-client widget usage, prompt reminders, event history, and Slack delivery

hiskudin:
## What

Four features, each independently useful. The through-line is the app's own premise: *you aren't watching the terminal*. Three of these close gaps where that stopped being true.

### 1. The widget shows Codex and Antigravity usage

The pill's gauge read `nav.quota` directly in eight places, so it could only ever show Claude — even though the Usage tab has carousel'd Claude / Codex / Antigravity since the Codex probe landed, and all three probes already run on the same poll tick.

The widget now **mirrors the Usage tab's selection**, so the pill and the tab can't disagree. `WidgetQuota` reduces whichever client is selected to the two rings and one countdown the pill can render: 5h/weekly for Claude and Codex; for Antigravity, which publishes neither, the model closest to its limit and the monthly prompt-credit pool.

Derived signals keep their exact old semantics — border colour, pulse rate and refresh cadence key off the short window alone; the mascot's stress signal off whichever window is tighter — so **a Claude-only user sees no change at all**.

Gemini is deliberately absent: it writes no rate-limit data to disk and serves no local endpoint, so a counter needs a live Code Assist API call. Filed as #163.

### 2. Reminders for unanswered prompts, and stalled sessions

A permission prompt fired exactly one banner. macOS slides that into Notification Center after a few seconds and nothing ever mentioned it again — so walking away turned a ten-second interruption into an agent blocked for its full 550s hook timeout.

Prompts still waiting now re-nudge on an interval (default 2 min, capped at three), carrying the same Allow/Deny buttons and a "Still waiting 4m" body. The menu bar shows a live count — the only trace left once a banner has expired.

**Reminders fire only for prompts we can prove are unanswered.** `notify.sh` creates the decision FIFO before emitting the event and removes it on exit, so the file existing means nobody has answered anywhere. Gemini and Antigravity get no FIFO (their hooks can't consume a decision) and are never reminded about: a reminder we can't verify would eventually fire for prompts already handled in the terminal, and a nudge that cries wolf is worse than none.

Watches are reconciled from observable state each tick rather than deregistered by each resolution path, so no future allow/deny route can forget to clean up.

Also flags sessions that report busy but have produced nothing past a configurable window (default off) — usually a wedged tool call. Sessions tab only, no banner: there's nothing to approve and a stall often clears itself.

### 3. Searchable, durable event history

`EventStore` is a live triage queue: capped at 100, pruned to 5 per session, gone on quit. Right for a list you act on, but "what did the agents do overnight" had no answer at all.

Adds an append-only JSONL log at `~/.stack-nudge/events.jsonl` (mode 0600, 30 days or 10k records, trimmed once per launch) and a History pane behind the Events tab, reachable with `H`, with a filter matching on message, agent and repo.

History is a **second pane rather than extra rows** on purpose. The queue is an action list — Enter approves, Backspace dismisses — and a replayed record has no FIFO to answer and no process to focus. Mixing them would make those keys mean different things depending on which row you landed on. The record is a slim projection of `NudgeEvent`, so inertness falls out of the schema rather than relying on a flag.

The log holds prompt and tool text, which the app previously kept in memory only — so it's documented plainly, written owner-only, disableable, and clearable from Settings.

### 4. Slack DMs when you're away from the desk

Notifications were entirely local. A permission prompt now also arrives as a DM **from StackNudge**, idle-gated by default (5 min, configurable to `Always` or up to an hour) so you don't get pinged about prompts you're looking at. Reminders skip the gate.

A **global mute deliberately does not suppress Slack** — mute means "not here", and Slack is the elsewhere. A per-session mute does. Titles only unless you opt into detail, since tool text can carry paths, hostnames and secrets, and this is the one path that leaves the machine.

Setup is one paste, and there is **no default app and no embedded credentials** — an embedded client ID would have meant anyone building this repo authorising StackOne's Slack app into their own workspace. The token lives in the Keychain; scripted provisioning can plant it in the config file, from where it's migrated and the line deleted.

**Route not taken, recorded so it isn't rediscovered:** Slack supports OAuth with PKCE, which authenticates a desktop app with no client secret and no server. It was built and it worked. But *"desktop redirects are not allowed to request bot scopes"*, so it only yields a **user** token; a user token can only post *as you*; and Slack never notifies you about your own messages — the DMs arrived silently. Slackbot reminders were the one remaining courier and Slack retired that API in March 2023. Hence a provisioned bot token.

## Also

`ConfigFile.write` now sets `0600`. An atomic write replaces the file, so the mode has to be reapplied on every write or it reverts to the umask default — which is how `~/.stack-nudge/config` came to be world-readable while carrying provisioning secrets.

## Testing

- `make build` and `make typecheck-tests` clean throughout.
- 5 new test files, ~90 cases: `WidgetQuotaTests`, `AttentionPolicyTests`, `EventLogTests`, `SlackCredentialsTests`, `SlackDirectoryTests`, `SlackDeliveryTests`.
- `swift test` can't run locally (Command Line Tools only, no Xcode), so each pure policy file was also compiled against a stub harness and the assertions executed: **30, 34, 26 and 55 passing** for the widget, attention, log and Slack layers respectively. CI runs the XCTest versions.
- Two defects were caught this way that `typecheck-tests` would not have: an unrealistic `xoxb-` fixture that failed the token length guard (present in the XCTest file too, so CI would have gone red), and a missing assertion that retention is *written back* rather than only filtered on read.
- Exercised live: Slack test message and a real permission event delivering end to end; the event log verified on disk at `0600`; config migration confirmed to scrub the token line.


panel/Panel.swift
```diff
@@ -2341,17 +2503,369 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+            .filter { !nav.isSessionMuted(event: $0.event, in: sessions.sessions) }
+
+        guard let watch = due.min(by: { $0.firstSeenAt < $1.firstSeenAt }) else { return }
+        let event = watch.event
+
+        let config = PanelConfig.load()
+        promptWatches[event.id]?.lastNudgedAt = now
+        promptWatches[event.id]?.remindersSent += 1
+
+        notifySlack(for: event, isReminder: true)
```
StuBehan, line 2825: Does this need to skip a snoozed prompt? We gate on the global mute and the per-session mute but not `snoozedUntil` - so S for 5 mins on a permission prompt still re-nudges at 2 and 4, banner + chime + slack DM. `PromptWatch.event` is the pre-snooze copy too so you'd have to look it back up in `store.events` 🤔

panel/MenuBar.swift
```diff
@@ -69,6 +90,19 @@ enum ConfigFile {
         return lines.joined(separator: "\n")
     }
 
+    /// Return `contents` with every uncommented assignment of `key` removed,
+    /// leaving surrounding lines and comments untouched.
+    static func strip(_ contents: String, key: String) -> String {
+        var lines = contents.components(separatedBy: "\n")
+        lines.removeAll { line in
+            let trimmed = line.trimmingCharacters(in: .whitespaces)
+            return !trimmed.hasPrefix("#") && trimmed.hasPrefix("\(key)=")
```
StuBehan, line 99: `parse` trims around the `=` but `strip` wants `KEY=` exactly - so a provisioned `STACKNUDGE_SLACK_BOT_TOKEN = xoxb-...` gets adopted into the keychain, scrubbed of nothing, and re-migrated every launch. Thats the line the README promises we delete. Match whatever `parse` accepts?

panel/Panel.swift
```diff
@@ -2341,17 +2503,369 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+            storeSlack(memberID: id, fromEmail: false)
+            nav.slackSetupNote = "Bot token and user stored"
+        case .unrecognised:
+            nav.slackSetupNote = pasted.isEmpty
+                ? "Clipboard is empty"
+                : "Clipboard isn't an xoxb- token or a U… member ID"
+        }
+    }
+
+    private func storeSlack(token: String) {
```
StuBehan, line 2555: We drop the Bool, and your comment on `store(botToken:)` says not to trust it on faith. Locked keychain means the add fails and the row still reads "Bot token stored" with test + notifications unlocked, then the test comes back "No Slack bot token" 🤷 Guard on it?

panel/PanelNav.swift
```diff
@@ -934,6 +1019,21 @@ final class PanelNav: ObservableObject {
...
         let rawCtx = Int(config["STACKNUDGE_CONTEXT_ALERT_THRESHOLD"] ?? "") ?? 0
         contextAlertThresholdK = Self.contextAlertThresholdOptions.min(by: { abs($0 - rawCtx) < abs($1 - rawCtx) }) ?? 0
+        eventHistoryEnabled = ConfigFile.bool(config, "STACKNUDGE_EVENT_HISTORY", default: true)
+        slackEnabled       = ConfigFile.bool(config, "STACKNUDGE_SLACK", default: false)
+        slackIncludeDetail = ConfigFile.bool(config, "STACKNUDGE_SLACK_DETAIL", default: false)
+        slackNotifyOnStop  = ConfigFile.bool(config, "STACKNUDGE_SLACK_STOP", default: false)
+        let rawSlackIdle = Int(config["STACKNUDGE_SLACK_IDLE_MIN"] ?? "") ?? 5
+        slackIdleMinutes = SlackDelivery.idleMinuteOptions
+            .min(by: { abs($0 - rawSlackIdle) < abs($1 - rawSlackIdle) }) ?? 5
+        slackTokenPresent = SlackCredentials.botToken() != nil
```
StuBehan, line 1029: Sync `SecItemCopyMatching`, and `loadFromConfig` runs from SettingsView's `.onAppear` - so we hit the keychain on main every time Settings opens. You moved this off main in `SlackNotifier.send` for exactly that reason, same here and in `startSlack`?

panel/Settings.swift
```diff
@@ -437,6 +451,30 @@ struct SettingsView: View {
...
+    private var slackReady: Bool {
+        nav.slackTokenPresent && nav.slackMemberID != nil
+    }
+
+    // The paste row doubles as the status line for the whole section: the last
+    // paste/lookup/test outcome wins, then a delivery error, then the resting
+    // state. A stale "token stored" next to a broken integration would be worse
+    // than saying nothing.
+    private var slackPasteValue: String {
+        if let note = nav.slackSetupNote { return note }
```
StuBehan, line 465: `slackSetupNote` never gets cleared and it wins over `slackError`, so "Test message sent" sits there forever and a revoked token never surfaces - which is the state the comment above says this is meant to prevent. Clear it on a failed delivery, or flip the precedence?

panel/PanelNav.swift
```diff
@@ -599,6 +616,26 @@ final class PanelNav: ObservableObject {
...
+    // carrying its own, so the pill and the tab can never disagree about whose
+    // numbers you're looking at. All three probes already run on the same tick
+    // (see runQuotaProbe), so switching client costs no extra I/O.
+    // Memoised because CompactView reads it from ~16 places per body pass and
+    // the pill re-renders at 10Hz while an agent is busy — recomputing meant
+    // rebuilding availableUsageClients every time. Invalidated by didSet on each
+    // of the four inputs below, so the cache can't outlive its sources.
+    private var widgetQuotaCache: WidgetQuota?
+
+    var widgetQuota: WidgetQuota {
```
StuBehan, line 629: Selection is an index into a list whose membership depends on which probes have data, so a claude hard-fail nils `nav.quota` and the pill quietly becomes Codex for a tick - rings, tooltip, border, countdown. Would holding the selected `UsageClient` rather than an index sort it? (the `.antigravity` comment in WidgetQuota is fighting the same thing from the other end)

panel/Panel.swift
```diff
@@ -2341,17 +2503,369 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+        // Deliberately not gated on nav.isMuted: mute means "stop interrupting
+        // me *here*", and Slack exists because the user is elsewhere — the same
+        // rule postBannerIfNeeded already applies to the first nudge. Gating the
+        // reminder too made the two inconsistent: muted users got the opening
+        // Slack DM and then silence. Local cues are still gated, below.
+        let localEligible = !nav.isMuted
+        let slackEligible = nav.slackEnabled && nav.slackTokenPresent && nav.slackMemberID != nil
+        // Nothing can be delivered, so don't spend a reminder from the budget —
+        // the prompt should still get its nudges once the mute lifts.
+        guard localEligible || slackEligible else { return }
```
StuBehan, line 2804: `localEligible` is just `!nav.isMuted`, so with banner, voice and sound all off and slack off we spend a reminder and deliver nothing. Fold the three config flags in? (menu bar count still updates so its minor)

panel/Panel.swift
```diff
@@ -2341,17 +2503,369 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+
+        let config = PanelConfig.load()
+        promptWatches[event.id]?.lastNudgedAt = now
+        promptWatches[event.id]?.remindersSent += 1
+
+        notifySlack(for: event, isReminder: true)
+
+        guard localEligible else { return }
+        if config.bannerEnabled {
+            postBanner(for: event,
```
StuBehan, line 2829: `postBanner` clears all delivered first, so a reminder for A wipes B's banner while B is still unanswered. Used to happen once per arriving event, now its every couple of minutes by default.

panel/Panel.swift
```diff
@@ -2341,17 +2503,369 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+
+    // MARK: - Event history
+
+    private let eventLog = EventLog()
+
+    // Load the durable log once at launch and attach it to the store if
+    // recording is on. Loading happens regardless of the toggle: turning
+    // recording off should stop new writes, not hide what's already there.
+    private func startEventHistory() {
+        nav.historyRecords = eventLog.load()
```
StuBehan, line 2673: On main in applicationDidFinishLaunching, and at the 10k cap thats a multi-MB read plus 10k decodes before the status item shows. Nothing needs it until someone presses H - background it, or defer to first open?

panel/EventLog.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    var projectName: String? {
+        guard let project, !project.isEmpty else { return nil }
+        return (project as NSString).lastPathComponent
+    }
+
+    // Free-text match across the fields a user would search by. Case-folded
+    // rather than lowercased so a Turkish locale can't make "I" stop matching.
+    func matches(_ needle: String) -> Bool {
+        let q = needle.folding(options: [.caseInsensitive, .diacriticInsensitive],
```
StuBehan, line 33: We re-fold the needle per record, so a keystroke at the 10k cap is ~50k folding calls, and `filteredHistory` being a computed prop means it re-runs on any nav change while the pane is open. Fold once at the call site? Also `q` - nothing else in the panel uses single letter names (same for the `e`/`d` below).

panel/EventLog.swift
```diff
@@ -0,0 +1,169 @@
...
+    let agent: String
+    let kind: String
+    let title: String
+    let message: String
+    let project: String?
+    let session: String?
+
+    // Stable within a load: the log is append-only and read in order, so the
+    // index is enough for SwiftUI identity without persisting a UUID per line.
+    var id: String { "\(at.timeIntervalSince1970)-\(agent)-\(message.hashValue)" }
```
StuBehan, line 23: Worth folding `project` and `session` in? Same agent kind, different repos, same tool text in the same second collides and ForEach moans about duplicate ids.

panel/EventStore.swift
```diff
@@ -189,6 +194,15 @@ final class EventStore: ObservableObject {
         // when many sessions are active in parallel.
         prune()
         if selectedID != event.id { selectedID = event.id }
+        // After the dedup check, so the log doesn't record a duplicate the
+        // panel deliberately dropped. The write itself is async.
+        log?.append(EventRecord(at: event.timestamp,
```
StuBehan, line 199: Written out again in `noteHistory` - miss one when you add a field and the pane and the file quietly disagree. `EventRecord(from: NudgeEvent)` and call it from both?

StuBehan (commented):
Solid PR, the fifo-as-proof-of-unanswered design is nice and the tests are proper tests. Checked the invariant holds - notify.sh creates the fifo before post_to_panel and the EXIT trap clears it, and the 550 matches promptLifetime.

Two I'd want sorted before merge, both inline - the snooze one and the config scrub one.

Not inline because they aren't in the diff:

PRIVACY.md still says "It transmits no data / It has no network access". The network line was already stale on main (updater, github auth) but this is the first thing sending event content off the box - agent, repo, and the tool text itself with SLACK_DETAIL on - plus the on-disk log. README covers it well, PRIVACY.md is what people read as the statement.

The History section's "Nothing is ever sent anywhere" sits a few paragraphs above the Slack section saying the same text can leave the Mac. Fine in context but I'd reword.

`promptLifetime = 550` and notify.sh's `local timeout=550` are the same number in two languages with nothing tying them - worth a comment from the shell side.

AskUserQuestion gets no fifo so no reminder and no menu bar count, even though the agent is blocked. Consistent with the rule so I don't think its wrong, but worth a README line next to the gemini/agy carve out?

And the `ConfigFile.persist` comment (and the PR description) says an atomic write reverts the mode on every write - I tested it, Darwin preserves an existing file's mode, 0600 in 0600 out. Only bites on creation. setAttributes is still right, just the reason isn't. Same for `EventLog.rewrite`, which I'd assumed was buggy for that and isn't.

PR #165, opened 2026-09-04
feat(agents): add pi coding-agent support

StuBehan:
Adds `pi` (`@earendil-works/pi-coding-agent`) as a first-class agent, alongside Claude Code / Codex / Gemini / Antigravity.

## Sessions
- Detect `pi` in the process scan (`SessionStore.detectAgent`), including the node-hosted `pi-coding-agent` bundle.
- New `PiTranscriptReader` parses pi's session JSONL (`~/.pi/agent/sessions/…/*.jsonl`) into the same `TranscriptStats` the Claude/Codex readers produce; `TranscriptStats.read` dispatches `/.pi/agent/sessions/` paths to it. Context occupancy = `input + cacheRead + cacheWrite`.
- `SessionStore.enrichPi` binds each pi session to its transcript from disk (matching the header `cwd`, not the folder encoding), so tokens/model show the moment the panel opens — no hook required. `Panel.refreshAllClaudeStats` keeps them live.

## Events
- pi has no Claude-style hooks, so `hooks/pi/stack-nudge.ts` is a global pi extension (auto-loaded from `~/.pi/agent/extensions/`). It forwards `agent_settled` (pi's Stop analog) to `notify.sh pi stop`, so pi gets the same banner/voice/handoff path as every other agent. `install.sh` installs it when `~/.pi/agent` is present.
- `notify.sh`: `pi` label; `walk_session_chain` seeds `AGENT_PID` from `NUDGE_AGENT_PID` (node-hosted agents show `comm=node`). pi is notification-only (no permission decisions), like Antigravity.

## tmux click-to-focus
pi renames its process title to `pi`, which clobbers the argv+env memory `ps -e` reads — so `TmuxFocus` could not read `TMUX_PANE` from pi's process to resolve its pane (it works for Claude, whose env stays readable). The extension writes a per-pid sidecar `~/.stack-nudge/pi-sessions/<pid>.json` with the pane/socket/host from inside the process, and `TmuxFocus.target` reads that first (falling back to the `ps` ancestry walk for other agents). Also adds a shell-pid fallback for events whose agent has already exited.

## Not included
- The macOS first-launch wizard (`Bootstrap.swift`) does not auto-wire pi yet — its model is JSON-hook-splicing and pi needs a file copy; `install.sh` wires it.
- No Usage-tab history/quota for pi (local models / own keys have no subscription window).

## Verification
- `swift build` green; test sources typecheck; new tests: `PiTranscriptReaderTests`, `TmuxFocusTests` sidecar coverage, a `detectAgent` case.
- Verified live on macOS: sessions show tokens/model, events fire with banner + voice, and Sessions/Events click-to-focus jumps to the tmux pane in iTerm2.

### DIFF
--- a/hooks/pi/stack-nudge.ts
+++ b/hooks/pi/stack-nudge.ts
@@ -1,0 +1,122 @@
+// stack-nudge integration for pi (@earendil-works/pi-coding-agent).
+//
+// pi has no Claude-Code-style hook system, but it does auto-load extensions
+// from ~/.pi/agent/extensions/*.ts in every session. This extension forwards
+// pi's lifecycle events to stack-nudge's installed hook entrypoint (notify.sh)
+// exactly as the Claude Code / Codex hooks do, so pi sessions get the same
+// banner, voice, attention-focus, and per-ticket handoff capture with no
+// stack-nudge-side special-casing (the panel keys everything off the `pi` agent
+// string, which flows through the same unix socket as every other agent).
+//
+// Install: copy to ~/.pi/agent/extensions/stack-nudge.ts (install.sh does this
+// automatically when ~/.pi/agent is present).
+
+import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
+import { spawn } from "node:child_process";
+import { mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { homedir } from "node:os";
+import { basename, join } from "node:path";
+
+// stack-nudge's installed hook entrypoint — the same script Claude/Codex invoke.
+const NOTIFY = join(homedir(), ".stack-nudge", "notify.sh");
+
+// Per-pid tmux sidecar. pi renames its process title to "pi", which clobbers the
+// argv+env memory region `ps -e` reads — so the panel CANNOT read pi's env from
+// outside to resolve its tmux pane (unlike claude, whose env stays readable).
+// The extension runs inside pi, so it's the only place that can read TMUX_PANE;
+// it writes the pane here for the panel's tmux focus to consume by pid. Analogous
+// to Claude Code's ~/.claude/sessions/<pid>.json sidecar.
+const SIDECAR_DIR = join(homedir(), ".stack-nudge", "pi-sessions");
+const sidecarPath = join(SIDECAR_DIR, `${process.pid}.json`);
+
+const removeTmuxSidecar = (): void => {
+  try {
+    rmSync(sidecarPath, { force: true });
+  } catch {
+    // ignore
+  }
+};
+
+const writeTmuxSidecar = (): void => {
+  const pane = process.env.TMUX_PANE;
+  // Not inside tmux: clear any sidecar this pid may have inherited from a dead
+  // tmux-hosted pi, rather than returning early and leaving it to point a paneless
+  // session at a stale pane. Keeps the write path idempotent for the pid either way.
+  if (!pane) {
+    removeTmuxSidecar();
+    return;
+  }
+  // TMUX is "<socket>,<serverPID>,<sessionN>"; the socket is the first field.
+  const socket = process.env.TMUX?.split(",")[0];
+  try {
+    mkdirSync(SIDECAR_DIR, { recursive: true });
+    writeFileSync(sidecarPath, JSON.stringify({
+      pane,
+      socket,
+      lcTerminal: process.env.LC_TERMINAL,
+    }));
+  } catch {
+    // never surface to the session
+  }
+};
+
+// pi's session id is the uuid half of "<timestamp>_<uuid>.jsonl". Deriving it
+// from the filename avoids reading the file just to echo the header id back.
+const sessionIdFromFile = (file: string | undefined): string | undefined => {
+  if (!file) return undefined;
+  const stem = basename(file, ".jsonl");
+  const underscore = stem.indexOf("_");
+  return underscore >= 0 ? stem.slice(underscore + 1) : stem;
+};
+
+// Fire the hook without ever blocking or throwing back into pi: a missing
+// notify.sh, a permission error, or a broken pipe must stay invisible to the
+// coding session. The payload mirrors the subset of the Claude hook JSON that
+// notify.sh reads (session_id, transcript_path, cwd); NUDGE_AGENT_PID seeds the
+// agent pid that walk_session_chain can't infer for a node-hosted agent.
+const fireHook = (event: string, sessionFile: string | undefined): void => {
+  const payload = JSON.stringify({
+    hook_event_name: event,
+    session_id: sessionIdFromFile(sessionFile),
+    transcript_path: sessionFile,
+    cwd: process.cwd(),
+  });
+  try {
+    const child = spawn("/bin/bash", [NOTIFY, "pi", event], {
+      stdio: ["pipe", "ignore", "ignore"],
+      env: { ...process.env, NUDGE_AGENT_PID: String(process.pid) },
+    });
+    child.on("error", () => {});
+    child.stdin?.on("error", () => {});
+    child.stdin?.end(payload);
+    child.unref();
+  } catch {
+    // never surface to the session
+  }
+};
+
+export default function (pi: ExtensionAPI): void {
+  // Write the tmux pane sidecar up front (the session is already running when
+  // the factory loads) and again on session_start, so the panel can focus the
+  // pane by pid from the first turn. Removed on shutdown to avoid a stale entry
+  // focusing the wrong pane if the pid is later reused.
+  writeTmuxSidecar();
+  pi.on("session_start", async () => writeTmuxSidecar());
+  pi.on("session_shutdown", async () => removeTmuxSidecar());
+
+  // agent_settled is pi's "won't continue on its own" signal — the analog of
+  // Claude Code's Stop hook. That's the moment the user's attention is wanted,
+  // and the point at which the handoff/token snapshot should be captured.
+  pi.on("agent_settled", async (_event, ctx) => {
+    if (!ctx.isIdle()) return;
+    // sessionManager is absent under `pi --no-session`; guard so the handler
+    // never throws (the event still fires, just without a transcript path).
+    let sessionFile: string | undefined;
+    try {
+      sessionFile = ctx.sessionManager?.getSessionFile();
+    } catch {
+      sessionFile = undefined;
+    }
+    fireHook("stop", sessionFile);
+  });
+}
--- a/install.sh
+++ b/install.sh
@@ -479,9 +479,27 @@ PY
   installed_any=true
 fi
 
+# pi (@earendil-works/pi-coding-agent). Unlike the others, pi has no hook-config
+# JSON to splice: it auto-loads extensions from ~/.pi/agent/extensions/*.ts, so
+# we install one that forwards pi's lifecycle events to notify.sh.
+if [[ -d "$HOME/.pi/agent" ]]; then
+  echo ""
+  echo "Detected pi (~/.pi/agent)"
+  PI_EXT_SRC="$SCRIPT_DIR/hooks/pi/stack-nudge.ts"
+  PI_EXT_DIR="$HOME/.pi/agent/extensions"
+  if [[ -f "$PI_EXT_SRC" ]]; then
+    mkdir -p "$PI_EXT_DIR"
+    cp "$PI_EXT_SRC" "$PI_EXT_DIR/stack-nudge.ts"
+    echo "  Installed extension    -> $PI_EXT_DIR/stack-nudge.ts"
+    installed_any=true
+  else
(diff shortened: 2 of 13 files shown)


panel/PiTranscriptStats.swift
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    // The first line of every pi session is a `session` header:
+    // {"type":"session","version":N,"id":"<uuid>","timestamp":"...","cwd":"..."}.
+    // Read only a bounded prefix — the header is tiny and the file can be MB.
+    private static func sessionHeader(path: String) -> (id: String, cwd: String)? {
+        guard let handle = FileHandle(forReadingAtPath: path) else { return nil }
+        defer { try? handle.close() }
+        let prefix = handle.readData(ofLength: 65536)
+        guard let text = String(data: prefix, encoding: .utf8) else { return nil }
```
hiskudin: **A pi session can silently never bind to its transcript.**

This decodes the *whole* 64KB prefix as UTF-8. If a multi-byte character straddles byte 65536, `String(data:encoding:)` returns nil and the header is lost — even though line 1 is perfectly intact and only ~110 bytes in.

Reproduced against this branch's code, with a 3-byte character placed at offset 65534:

```
64KB prefix decodes as UTF-8: false
header on line 1 is intact:   true
locate() result:              nil  <-- session NOT bound
read() result:                240 tokens   (unaffected — whole-file decode)
```

`enrichPi` then skips the session: no session id, no `lastActivityAt`, no context stats — permanently, and with nothing logged. Any transcript containing an emoji, an accented name, or a smart quote can land on this; it only takes one falling on the boundary, which for a 64KB window is not a rare event across a session history.

Decoding only the first line fixes it, and is what the surrounding comment already describes ("the header is tiny"):

```swift
let prefix = handle.readData(ofLength: 65536)
guard let newline = prefix.firstIndex(of: 0x0A),
      let text = String(data: prefix[..<newline], encoding: .utf8)
else { return nil }
```

That also removes the need for the `split(separator:maxSplits:)` below.

hiskudin (commented):
Read the whole diff, built the branch (`make build` + `make typecheck-tests` both clean), and exercised `PiTranscriptReader` against real files with a standalone harness.

**The design is right.** Matching on the `cwd` stamped in the session header rather than reconstructing pi's folder encoding is the correct call — that's the part that would otherwise break on every pi release. `enrichPi` running inside `discover()` keeps it off the main thread, the `NUDGE_AGENT_PID` seeding correctly guards against being overwritten by the walk (`[[ -z "$AGENT_PID" ]]`), and the extension's error handling genuinely can't surface into a coding session.

Three things I'd fix, one of which I was able to reproduce. Details inline.

### Also, briefly

**Two independent derivations of the pi session id.** The extension derives it from the filename (`<timestamp>_<uuid>.jsonl`, everything after the first `_`); `PiTranscriptReader.sessionHeader` reads `id` out of the header JSON. Nothing asserts these agree, and `test_locate_matchesByHeaderCwd_newestWins` uses `old.jsonl` / `new.jsonl` — so the real filename format is never exercised. If they ever diverge, event↔session correlation breaks silently and the symptom (stats not updating after a turn) points nowhere near the cause. I couldn't check, as pi isn't installed on my machine — but if the header id *is* always the filename uuid, a single assertion in the reader tests using a realistically-named fixture would pin it.

**`locate()` cost.** It's called per pi session on every poll (3s foreground), and each call lists every session directory, stats every `.jsonl` in each, and reads 64KB from the newest of each. Fine at a handful of projects, less so at fifty. `CodexQuotaProbe` in this repo caches on `path|size|mtime` for exactly this reason and would port over cheaply.

panel/TmuxFocus.swift
```diff
@@ -25,18 +25,72 @@ enum TmuxFocus {
...
+    // the same pane still does. Every ancestor up to the tmux server belongs to
+    // the same pane, so the first one carrying TMUX_PANE names the right pane.
+    // Bounded so an unexpected/detached tree can't loop. Runs on a background
+    // queue (callers dispatch); each `ps` has a timeout so a hang can't wedge us.
     static func target(agentPID: Int) -> Target? {
+        // Sidecar first: agents whose env `ps` can't read (pi renames its process
+        // title, clobbering the argv+env region) write their tmux pane to a
+        // per-pid sidecar from inside the process. When present it's authoritative
+        // and needs no `ps` at all. claude et al. have no sidecar and fall through.
+        if let resolved = sidecarTarget(pid: agentPID) {
```
hiskudin: **A stale sidecar can hijack focus for a different agent.**

This is consulted first, for **every** agent, and `sidecarTarget` does no staleness check — no age, no liveness, no check that the pid is even pi.

The extension removes the sidecar on `session_shutdown`, and its comment says that's to avoid exactly this. But that only covers graceful exit: SIGKILL, a closed terminal window, or a crash leaves `~/.stack-nudge/pi-sessions/<pid>.json` behind. When the OS later reuses that pid — for Claude Code, say — clicking that session focuses **pi's old pane**, and the Claude pane never gets focus. The failure is silent and looks like "click-to-focus is flaky".

This repo already solved the same class of problem: `SessionStore.sweepStaleClaudeSidecars` exists because *"Claude Code writes these but doesn't garbage-collect them"*, with PID-liveness and minimum-age guards. Either reuse that pattern for `pi-sessions/`, or make the sidecar self-validating — recording the process start time and comparing against `ps -o lstart` on read would make a reused pid unambiguous.

Cheapest partial mitigation if you'd rather not do either: only consult the sidecar when the session's agent is `pi`, which bounds the blast radius to pi-on-pi reuse.

install.sh
```diff
@@ -479,9 +479,27 @@ PY
...
+# we install one that forwards pi's lifecycle events to notify.sh.
+if [[ -d "$HOME/.pi/agent" ]]; then
+  echo ""
+  echo "Detected pi (~/.pi/agent)"
+  PI_EXT_SRC="$SCRIPT_DIR/hooks/pi/stack-nudge.ts"
+  PI_EXT_DIR="$HOME/.pi/agent/extensions"
+  if [[ -f "$PI_EXT_SRC" ]]; then
+    mkdir -p "$PI_EXT_DIR"
+    cp "$PI_EXT_SRC" "$PI_EXT_DIR/stack-nudge.ts"
+    echo "  Installed extension    -> $PI_EXT_DIR/stack-nudge.ts"
```
hiskudin, line 493: **`install.sh` grew a step that `uninstall.sh` didn't.**

`uninstall.sh` splices the hook entries back out of `~/.claude`, `~/.gemini`, and `~/.gemini/antigravity-cli`, but has nothing for pi — so `~/.pi/agent/extensions/stack-nudge.ts` survives an uninstall and keeps spawning a now-missing `notify.sh` on every pi turn.

It fails silently (the extension's `try`/`catch` and `child.on("error")` see to that), so this is leftover cruft rather than breakage — but it's cruft that runs in every pi session forever, and the uninstall flow is advertised in the README as complete. A `rm -f "$HOME/.pi/agent/extensions/stack-nudge.ts"` alongside the others closes it.

Worth also clearing `~/.stack-nudge/pi-sessions/` if that directory isn't already covered by the wholesale `~/.stack-nudge` removal.

panel/TmuxFocus.swift
```diff
@@ -25,18 +25,72 @@ enum TmuxFocus {
...
+            debug("target(pid=\(agentPID)) -> pane=\(resolved.pane) via sidecar "
+                + "socket=\(resolved.socket ?? "default") host=\(resolved.hostBundleID ?? "nil")")
+            return resolved
+        }
+
+        var pid = agentPID
```
hiskudin: The ancestry walk changes behaviour for **every** agent, not just pi.

Previously a failed env read on the agent pid returned nil and focus simply didn't happen. Now it climbs up to 10 ancestors and may resolve a pane. The reasoning in the comment is sound — ancestors up to the tmux server share a pane — but it's worth a sanity check that there's no path where an agent inherits a *different* pane's `TMUX_PANE` from an ancestor (a process launched via `tmux send-keys` from another pane, or re-parented after its shell exited). Trading "no focus" for "focus the wrong pane" would be a regression for agents that already worked.

Minor: worst case this is 20 `ps` spawns (a `resolveOne` + a `parentPID` per level) on a click. Fine in practice, but `ps -o ppid=,command=` in one call per level would halve it.

hooks/pi/stack-nudge.ts
```diff
@@ -0,0 +1,116 @@
...
+// outside to resolve its tmux pane (unlike claude, whose env stays readable).
+// The extension runs inside pi, so it's the only place that can read TMUX_PANE;
+// it writes the pane here for the panel's tmux focus to consume by pid. Analogous
+// to Claude Code's ~/.claude/sessions/<pid>.json sidecar.
+const SIDECAR_DIR = join(homedir(), ".stack-nudge", "pi-sessions");
+const sidecarPath = join(SIDECAR_DIR, `${process.pid}.json`);
+
+const writeTmuxSidecar = (): void => {
+  const pane = process.env.TMUX_PANE;
+  if (!pane) return; // not inside tmux — nothing for the pane-focus path to use
```
hiskudin: Nit, but it feeds the stale-sidecar problem: when pi isn't running under tmux this returns early **without** clearing any pre-existing `<pid>.json`. So a pid inherited from a dead tmux-hosted pi keeps its stale sidecar, and this session — which has no pane at all — would resolve to that old pane.

`removeTmuxSidecar()` before the early return makes the write path idempotent for the pid either way.

panel/PiTranscriptStats.swift
```diff
@@ -0,0 +1,156 @@
...
+    static func locate(cwd: String, root: String = sessionsRoot) -> PiSessionRef? {
+        let fileManager = FileManager.default
+        let cacheKey = "\(root)::\(cwd)"
+
+        // Fast path: re-scan only the remembered directory, still verifying the
+        // header cwd so a reused/rotated folder can never return a wrong match.
+        dirCacheLock.lock()
+        let cachedDir = dirCache[cacheKey]
+        dirCacheLock.unlock()
+        if let cachedDir,
```
hiskudin: **The fast path pins to a directory and never reconsiders it.**

Once `dirCache["root::cwd"]` is set, `locate` returns that directory's newest transcript for as long as its header `cwd` still matches — so a *newer* session for the same cwd in a *different* directory is never picked up. The cache is static, so this persists for the life of the process.

Reproduced against this branch:

```
cold locate       -> OLD   (expect OLD)
after new session -> OLD   (expect NEW)
STALE: fast path pinned to the old directory; the live session is never picked up
```

The scenario is the one this file already guards against elsewhere: `locate`'s own comment says pi's folder wrap rule "has drifted across versions", which is precisely why it matches on the header `cwd` rather than reconstructing the name. A pi upgrade that changes the wrap for an existing project leaves the panel showing that project's *previous* session — stale tokens, stale session id — until the app restarts.

`test_locate_matchesByHeaderCwd_newestWins` still passes because it only ever runs against a cold cache. The warm path has no coverage.

**A cheap correct invalidation:** a new per-cwd folder bumps the *root* directory's mtime, and appends inside an existing subdirectory don't. I verified both:

```
new subdir bumps root mtime:      true
write inside a subdir bumps root: false
```

So caching `(rootMtime, dir)` and falling back to the full scan when root's mtime has moved keeps the saving in the steady state — appending to today's transcript, which is the common case — while staying correct when a directory appears. Worth a test that goes through `locate` twice with a new directory in between, since that's the case the cache exists to skip.

Entirely my doing for raising the cost as a nit without suggesting how to invalidate it — happy for the simplest resolution to be dropping the cache. It was a "lower" item and the full scan was never the bug.

hiskudin (commented):
Re-reviewed. The three findings are fixed, and two of the fixes are better than what I suggested:

- **The 64 KiB header decode** now slices at the first newline and hands the bytes straight to `JSONSerialization`, skipping the `String` round-trip entirely. Cleaner than the fix I proposed, and the regression test places the `€` so its first byte lands exactly on the read boundary — that's the right fixture.
- **The stale sidecar** is gated on `isLivePi`. Using `comm` (readable even though pi's env isn't) to confirm the pid is still a live `pi` is a neater answer than the age/liveness sweep I pointed at, and it closes the reuse case at the source rather than expiring around it.
- **`uninstall.sh`** now removes the extension, and the ancestry walk is gone — which also retires my "this changes behaviour for every agent" note, since the sidecar was already the authoritative path for pi and the walk was belt-and-braces.

The `test_locate_headerSessionIDMatchesFilenameDerivation` test is exactly what I was asking for on the two independent session-id derivations, with a realistic filename. Good.

**One new problem, and it came from my own nit** — I raised `locate()`'s per-poll cost as a "lower" item and the `dirCache` added in response has a correctness regression. Detail inline. Everything else here is ready; this is the only thing holding an approval.

hiskudin (approved):

PR #167, opened 2026-09-07
feat(panel): navigable history rows, and an events footer that fits the panel

StuBehan:
The events footer has been running off the panel since 1.31.0. The history PR added a `History` hint, which made it eight hints in the live queue - 688pt of them against a 600pt panel - so the bar ran out past the leading edge and took `Open editor` with it.

`FooterHint` is `.fixedSize()` inside `HStack { Spacer(); hints }`, so there was never a shrink path. It was already 16pt over in 1.30.0 when muted with a Stop row selected, which is why one more hint tipped it into every state rather than just that corner.

### The bar

`ShedToFitLayout` replaces the HStack in `PageFooter`. It's a `Layout` (the macOS 13 API `FlowLayout` already uses) that measures the hints up front, drops the lowest `footerShedOrder` ones until the rest fit, and trailing-aligns what's left. Shed hints get parked off-bar and the footer clips, so nothing can escape the panel again. The `PageFooter { ... }` ViewBuilder API doesn't change, so the other nine footers get the same protection for free - Sessions and the Usage sessions pane are both sitting on 6-10pt of headroom at the 560pt minimum, so they were next in line anyway.

Then the events bar earns its width rather than shedding: `⌘↑↓` rides along on the `Select` hint as a second keycap instead of paying for a `Top/Bottom` label of its own. That's the whole saving, and it means nothing gets shed at the default width - 590pt worst case (a Stop row selected while muted) against the 600pt default, with every shortcut still advertised. Better than 1.30.0 managed, which was 608pt and already over. At the 560pt minimum it gives up `Snooze`, which is dimmed on most rows anyway.

The events bar is data now, `PanelContentView.eventsFooterHints`, same shape as `settingsRows`. Other pages keep their separate `Top/Bottom` hint - they have the room.

### History, from the keyboard

The pane was read-only *and* mouse-only, which on a 30-day log is a bit useless. `↑↓` move a row selection and `⌘↑↓` jump to newest/oldest, the same as the live queue and Sessions. Selected row gets the same accent tint as `EventRow` and the viewport follows it. Selection moves over the *filtered* rows, so `⌘↓` lands on the last visible row rather than the last recorded one. `⌘↑↓` in the history pane also used to move the live queue's hidden selection, which it no longer does.

`H` gets you in, `Esc` gets you out, and that's the lot. The footer was briefly claiming both a `Back` and a `Hide` in there, which was a lie: `Esc` steps back to the live queue from this pane, it doesn't hide the panel. One key, one meaning, one hint.

The pane also opens with the filter field unfocused now (it used to grab it), so the rows have the keyboard rather than the caret. `/` or any printable character hands it over, and `⏎` hands it back without clearing what you typed - otherwise typing a filter left the field owning `↑↓` and the results you'd just narrowed down were unreachable.

Two things worth flagging from that:

- AppKit select-alls a field's contents when it becomes first responder, so seeding the filter with the character you typed got wiped by the next keystroke - type "eng", get "ng". The collapse to a caret at the end hangs off `onChange(of: historyFieldFocused)`, ie off focus actually landing rather than off a predicted runloop hop. Side effect is that `/` on an existing filter extends it now rather than replacing it, which feels like the better default for a box that Esc already clears.
- `isFilterInput` was testing for "not a control character". AppKit reports arrows and function keys as private-use codepoints from U+F700 up, and those are neither control nor illegal, so an arrow key repeat put an invisible U+F703 in the filter *and* took the field's focus. It tests positively now. That one only turned up from driving the app, not from the tests, so the history key routing is a pure `historyKeyAction(keyCode:characters:filterIsEmpty:)` function with assertions on every key.

### `make test-without-xcode`

Second half of this PR, and it started as review feedback: `swift test` needs the XCTest module, which only ships with full Xcode, so on a Command Line Tools-only machine the test sources never get *run* locally at all. Everyone hits it and everyone hand-rolls their own worse harness.

`scripts/typecheck-tests.sh` already solved half of it (compile the tests in-module with stand-ins from `scripts/xctest-shim.swift`), and the stand-ins were deliberately inert because type-checking never executes them. So: make them evaluate and record, and add `scripts/run-tests-without-xcode.sh` to compile and run. One shim, two consumers, and the type-check path is untouched because it passes `-typecheck`.

```
make test-without-xcode                       # 600 test(s), 1621 assertion(s), 0 failure(s)
make test-without-xcode NAME=HistoryKeyAction # just the matching Class.method
```

Discovery is pattern-based, so it asserts its own count against the declared `func test` count and fails hard on a mismatch - my first cut silently skipped 21 `throws` tests, which is exactly the sort of quietly-smaller green number that means nothing. It's a fast local loop, not a replacement: serial, one process, no isolation or timeouts. `make test` and CI stay the authority.

It has already earned its keep twice. It caught `↓` with nothing selected skipping the top row, because I'd copied `EventStore`'s `firstIndex { ... } ?? 0` idiom, which conflates "nothing selected" with "row 0". (Note `EventStore.selectNext` has that same latent skip in the live queue - it just rarely shows because something is usually selected there already. Left alone, happy to do it separately.) And a hand-copied assertion set drifted from the real test file earlier on this PR and cost a red CI run, which is what pushed me to run the real sources instead.

### Testing

`FooterHintTests.swift` covers the shed decision, the hint set per pane and per selection state, history key routing, and history selection movement.

Two of those came out of review:

- The shed-order comparator is a pure `ShedToFitLayout.shedSequence(of:)` shared by `makeCache` and the tests, rather than the tests restating the rule and passing whatever `makeCache` happened to do.
- `FooterShedOrderPropagationTests` runs a real layout pass over `FooterHintRow`s and asserts the annotation arrives at the parent `Layout`. That hop (spec → the modifier inside `FooterHintRow`'s body → `LayoutValueKey`) is the one part of the mechanism resting on SwiftUI's semantics rather than arithmetic, and a silent break there degrades shedding to trailing-most-first: bar still fits, screenshots still look right, wrong hint dropped.

Builds clean for arm64 and x86_64, and I rendered every footer state to PNG to check the layout by eye since the original complaint was visual.

Titled `feat` rather than `fix` because the history row navigation is new behaviour, so it wants a minor bump and a Features line in the changelog. Shout if you'd rather it went out as a patch.


Tests/StackNudgePanelCoreTests/FooterHintTests.swift
```diff
@@ -0,0 +1,340 @@
...
+
+// The Events bar carries more hints than any other page, and #164 pushed it past
+// the panel width — FooterHint is .fixedSize() inside a trailing-aligned row, so
+// there was no shrink path and the leading hints ran off the panel. These pin the
+// two halves of the fix: the shed decision itself, and the hint set the Events
+// tab declares (including which entries are the sheddable ones).
+final class ShedToFitLayoutTests: XCTestCase {
+
+    // Hints 0…n-1 each 100pt wide, so widths in the assertions read as counts.
+    private func cache(_ shedOrders: [Int?], width: CGFloat = 100) -> ShedToFitLayout.Cache {
```
hiskudin, line 17: **This re-implements the code under test.**

The helper builds its `Cache` by duplicating `makeCache`'s comparator verbatim:

```swift
let order = shedOrders.indices.sorted {
    let lhs = shedOrders[$0] ?? .max
    let rhs = shedOrders[$1] ?? .max
    return lhs == rhs ? $0 > $1 : lhs < rhs
}
```

So `makeCache` is never exercised, and neither is the chain that actually carries the annotation: `FooterHintSpec.shedOrder` → `FooterHintRow`'s `.footerShedOrder()` → `subviews[i][FooterShedOrderKey.self]`.

I checked that chain works — a `LayoutValueKey` set inside a wrapper view's body does reach the parent `Layout`, verified with a real hosting-view layout pass. So this isn't a bug today. But it's the one link in the shed mechanism that depends on SwiftUI's semantics rather than pure arithmetic, and nothing here would notice it breaking: wrap `FooterHintRow` in a container, or move the modifier, and every hint reads `.max`. Shedding then degrades to trailing-most-first — the bar still fits, the PNGs still look right, and you quietly lose `Hide` before `Snooze`.

Feeding `makeCache` the real subviews would cover both at once. Failing that, at minimum have the helper call `makeCache`'s comparator rather than restating it — the PR body notes a hand-copied assertion set already drifted once and cost a red CI run, and this is the same hazard in a smaller place.

panel/Panel.swift
```diff
@@ -276,12 +277,37 @@ struct PanelContentView: View {
...
+        .onChange(of: nav.historyFilterFocusRequests) { _ in
+            historyFieldFocused = true
+            // AppKit selects a field's entire contents when it becomes first
+            // responder, which would throw away the character that asked for
+            // focus in the first place (type "eng", get "ng"). Collapse to a
+            // caret at the end once focus has landed — measured to take one
+            // runloop hop. It also makes / on an existing filter extend it
+            // rather than replace it, which is the more useful default for a
+            // filter box that Esc already clears.
+            DispatchQueue.main.async { Self.collapseFilterSelectionToEnd() }
```
hiskudin: **A timing assumption whose failure mode is the bug it fixes.**

One `main.async` hop, then reach into `NSApp.keyWindow?.firstResponder as? NSTextView`. The comment says a miss is harmless because "the seeded character is overwritten by the next keystroke" — but that's exactly the "type `eng`, get `ng`" behaviour being fixed. So a miss doesn't fail safe, it reintroduces the bug intermittently, and intermittently is the hard kind to chase.

"Measured to take one runloop hop" is true of what you measured, but it's SwiftUI's focus scheduling rather than a documented contract — it can move with an OS update, and it's the sort of thing that behaves differently under load or with the field already focused.

Binding the collapse to focus actually landing removes the guess entirely:

```swift
.onChange(of: historyFieldFocused) { focused in
    if focused { Self.collapseFilterSelectionToEnd() }
}
```

That fires when the state you care about is true, rather than when you predict it will be.

panel/Panel.swift
```diff
@@ -276,12 +277,37 @@ struct PanelContentView: View {
...
+        }
+    }
+
+    // The history filter is the only field focused at this point, so the key
+    // window's field editor is it. A miss is harmless: the seeded character is
+    // overwritten by the next keystroke, which is the pre-collapse behaviour.
+    private static func collapseFilterSelectionToEnd() {
+        let key = NSApp.keyWindow ?? NSApp.windows.first(where: { $0.isKeyWindow })
```
hiskudin: Nit: the fallback is unreachable. `NSApp.keyWindow` *is* the window whose `isKeyWindow` is true — if the first expression is nil there is no such window for the second to find, so `?? NSApp.windows.first(where: { $0.isKeyWindow })` can never contribute. `NSApp.keyWindow` alone says the same thing.

hiskudin (commented):
Read the whole diff, built the branch (`make build` + `make typecheck-tests` clean, CI green), and ran two of my suspicions against the real code rather than reporting them.

**Both turned out to be wrong**, which is worth recording since it's the interesting part of the review:

- I thought `filterInput` might not exclude AppKit's private-use codepoints, since `CharacterSet.symbols` covers `Sm/Sc/Sk/So` and I wasn't sure where `Co` landed. Ran U+F700–U+F739 and every control character through the real expression: all correctly rejected, all printable input (including CJK, combining accents and emoji) accepted. The positive test is right.
- I thought `FooterHintRow` applying `.footerShedOrder()` *inside its own body* might not reach `ShedToFitLayout`, which would silently degrade shedding to trailing-most-first — bar still fits, PNGs still look right, wrong hint dropped. Built a minimal `NSHostingView` layout pass with the same wrapper shape: the values arrive intact (`[7, 2, 5]`). It propagates.

So: **no defects found.** The `keptIndices` logic holds up on the edges I could construct (empty cache, single hint, exact fit, can't-shed-enough), `FooterHint`'s own `.padding(.leading, 10)` means dropping the `HStack` spacing was safe, the history guard is still above the `S` branch so #164's snooze-while-typing fix survives, `.onExitCommand` is preserved so Esc still works while the field holds focus, and `jumpToEdge` branching per pane is the right fix for ⌘↑↓ moving the live queue's hidden selection.

Three things worth raising, none blocking. Details inline, plus one general note:

**The 1616-assertion figure isn't reproducible from the PR.** The XCTest shim isn't in the diff, so nobody else can run it. That's fine as far as correctness goes — CI runs real `swift test` under Xcode and it's green — but the number reads with more authority than a reviewer can actually check. Given `swift test` genuinely can't run without full Xcode (I've been hand-rolling one-off harnesses for the same reason all week), a committed shim would be worth more than the figure: everyone gets the fast loop, and it stops each of us reinventing a worse one.

hiskudin (approved):
All three points addressed, and the shim is a better outcome than I asked for.

- **`shedSequence(of:)`** extracted and shared, so the tests now assert the production comparator instead of a restatement of it — and it gets a direct test of its own.
- **`FooterShedOrderPropagationTests`** closes the gap exactly: a real `NSHostingView` layout pass reading the annotation back off `FooterHintRow`. That's the one hop resting on SwiftUI semantics, and it's now pinned. (I'd verified it works with the same technique before commenting; good to have it in the suite rather than in my scratch directory.)
- **The caret collapse** keys off focus actually becoming true. The comment gets the reasoning right — a miss didn't fail safe, it reintroduced the bug intermittently.
- Dead `keyWindow` fallback gone.

**On the shim** — committing it was the right call, and I ran it rather than taking the numbers on faith:

```
600 test(s), 1621 assertion(s), 0 failure(s)
```

I also injected a deliberately wrong expectation to confirm it can actually fail, and it named the right test:

```
FAIL  WidgetQuotaTests.test_claude_mapsFiveHourAndSevenDay
600 test(s), 1621 assertion(s), 1 failure(s)
```

The discovery self-check earns its keep: it catches the likely failure (awk losing the enclosing class, so `cls == ""` and the method is skipped while `grep` still counts it). Worth knowing its limit — both counters use the same `^    func test` pattern, so a method that *neither* matches is invisible to the check rather than caught by it. Not worth changing; worth knowing.

One latent gap, no action needed now: the generated runner calls `setUp`/`tearDown` but not `setUpWithError`/`tearDownWithError`, which the shim declares. Nothing in the suite overrides them today, so a test that started to would silently run without its fixture. A `fatalError` in those two stand-ins would turn that into a loud failure the day it matters.

Approving.

PR #169, opened 2026-09-08
fix(panel): read tmux pane titles, and an opt-in to name sessions from them

hiskudin:
## What

Two related pieces, both about terminal tab titles.

**1. tmux pane titles are now read.** Tab titles were already collected for iTerm2, Terminal.app and VS Code/Cursor, but `TmuxIntegration` only ever set a pane id — so tmux rows showed the terminal and the path with nothing in between. `#{pane_title}` is the same OSC signal iTerm2 exposes as `autoName`, and it's what Claude Code rewrites each turn (`✳ Review repository structure`), so it's exactly the label those rows were missing.

**2. A tab title can now name a session** — behind `Settings → Name from tab titles` (`STACKNUDGE_TAB_TITLE_NAMES`, **default off**).

Warp and Ghostty are out of scope here.

## Why the naming half is opt-in

`SessionLabel`'s header already explained why the tab title was excluded outright: on iTerm2 a manual rename, an OSC escape and the profile name all write the same `autoName`, and Claude Code overwrites it every turn — so there's no way to read *"the user chose this"* back out. tmux is no better; `#{pane_title}` is the same OSC signal.

That argument is sound for a default, not for a prohibition. Plenty of people title tabs deliberately and want that name back. So it's a choice, off unless asked for, and it sits **below** both signals that *are* deliberate — a rename in the Sessions pane and a name set inside the agent still win. It only ever displaces the fallback to the project folder.

When on it applies everywhere a session is named: Sessions rows, the compact widget, event rows, banner titles, spoken nudges, and Slack DMs. All six route through two `SessionLabel` functions, so no surface is left inconsistent.

## Design notes

**No cache**, unlike the AppleScript integrations' 15s. Measured on this machine: `list-panes` at **8.4ms** vs osascript at **90ms** (and that was iTerm2's *not-running* fast path). At the 3s poll cadence the query costs less than the staleness would — and a cached title would name the previous turn's task once it can reach a banner.

**One query per distinct server.** Pane ids (`%4`) are only unique within a tmux server, so titles can't be pooled into one flat map.

**`AppActivator.tmuxPath()` / `tmuxEnv()` go private → internal** rather than being duplicated. The locale half is load-bearing, not hygiene — see below.

## Two things testing caught

**The locale fix is required.** Running the real command under `env -i` (as a launchd-spawned panel would), Claude's `✳` came back as literal underscores:

```
%   0   _   _       R   e   v   i   e   w       r   e   p   o ...
```

This is the same bug the focus path already hit once and documented; reusing `tmuxEnv()` is what avoids repeating it.

**A real bug in the first cut of this PR.** I compiled the actual `TmuxIntegration` against stubs and ran it on a live tmux server. `hostNames()` returned `hiskiass-macbook-pro-2.local` — lowercase — while tmux seeds pane titles with `Hiskiass-MacBook-Pro-2.local`. `ProcessInfo.hostName` gives the mDNS spelling; tmux uses the SystemConfiguration one. **The sentinel filter never fired**, so every untitled pane would have been named after the machine — and with the toggle on, that name would have titled a Slack DM.

The unit tests passed only because I'd written same-case fixtures. Now compared case-insensitively, with a regression test using the real mismatched casing. Verified end-to-end on a scratch server:

```
raw tmux:  %0 -> «Hiskiass-MacBook-Pro-2.local»   %1 -> «deploy pipeline»
real code: %1 -> «deploy pipeline»                 (%0 correctly filtered)
```

## Deliberate gap

**A renamed tmux *window* isn't picked up.** The signal exists — `show-options -w -t <win> automatic-rename` reads back `off` only after a manual rename — but it costs a subprocess per window, and the `#{automatic_rename}` *format* that would give it away for free renders empty whether or not the window was renamed (verified both ways). Batching is the rule the `TerminalIntegration` protocol exists to enforce, so this is a comment rather than N more spawns per poll.

## Testing

- `./build.sh` clean
- `make test-without-xcode` → **634 tests, 1683 assertions, 0 failures** (22 new)
- Live-path verification via a standalone harness compiling the real `TmuxIntegration` against stubs, run under `env -i` against a real tmux server

## Since review

- Trim the spinner glyph Claude cycles into the tab title (`✳`/`✻`/`✽`/`·`) — it spans two Unicode categories, so a category-only rule still let the name churn frame by frame.
- A Slack DM now requires `STACKNUDGE_SLACK_DETAIL` before it will use a tab title, since that switch is already the gate for local text on the one path that leaves the machine, and prompt frameworks put `user@host: /full/path` in tab titles.
- `STACKNUDGE_TAB_TITLE_NAMES` documented in `notify.conf.example`.
- Test coverage extended to the query contract itself after mutation testing showed five mutations — dropping `#{host}`, changing the delimiter, dropping `-a`, pooling servers, dropping the UTF-8 locale — all left the suite green.


panel/SessionLabel.swift
```diff
@@ -47,17 +56,46 @@ enum SessionLabel {
...
+    // no business filtering what someone typed into a tab.
+    //
+    // VS Code and its forks are excluded, because their `tabName` is not a tab
+    // title at all — VSCodeIntegration fills it from the OS *window* title that
+    // notify.sh captures ("Panel.swift — stackone — Cursor"), which names the
+    // file you happen to be looking at and changes every time you switch tabs in
+    // the editor. That is fine for the meta row it was built for, and unusable
+    // as a name: it would churn, and it would be read aloud em-dashes and all.
+    // The window title stays visible in the row; it just can't title a Slack DM.
+    private static func tabTitle(of session: Session) -> String? {
```
StuBehan, line 93: with the toggle on, does `✳ Review repository structure` end up going to `say`? I had a go and `say "✳"` gives byte-identical audio to `say "eight spoked asterisk"` - 1.23 seconds of it! So the nudge becomes "still waiting on eight spoked asterisk Review repository structure" 😬

`expandForSpeech` only splits on spaces so the glyph survives as its own word, and the reminder path at Panel.swift:2990 doesn't go through `expandForSpeech` at all. The comment right above here says a VS Code window title "would be read aloud em-dashes and all" - feels like thats the same problem, and excluding VS Code only fixes half of it? Claude's own title is the example in the header.

(not tmux specific either - iTerm2 was already filling `tabName`, so this turns on for every terminal)

panel/Panel.swift
```diff
@@ -2580,15 +2603,17 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
         else { return event.title }
         return "\(event.title) — \(label)"
     }
 
     private func sessionLabel(for event: NudgeEvent) -> String? {
         SessionLabel.chosenName(for: event,
                                 in: sessions.sessions,
-                                persistence: SessionPersistence.shared)
+                                persistence: SessionPersistence.shared,
+                                allowTabTitle: nav.tabTitleNames)
```
StuBehan, line 2616: this is the Slack and voice leg, and the comment on `SlackDelivery.text` says detail is opt-in because tool text "can carry paths, hostnames, and secrets in command lines, and this is the one path that leaves the machine".

`label` used to only ever be a name someone typed or the Claude sidecar name. Now its whatever the program in the pane wrote, and it goes into `subject` without `slackIncludeDetail` getting a say. Default zsh is fine (I checked, title just stays at the host sentinel after a `cd`) but anyone with a precmd title hook - omz, starship, most dotfiles - is putting `user@host: /full/path` in there.

Should the tab title be gated on `slackIncludeDetail` for the Slack leg specifically? Or at minimum called out in the README, because someone who left detail off on purpose is now getting local text out through a different door.

panel/SessionLabel.swift
```diff
@@ -13,19 +13,28 @@ import Foundation
...
+// OSC title escape and the profile name all write the same `autoName` variable,
+// so there is no way to read "the user chose this" back out — and Claude Code
+// rewrites the tab title every turn, so a manual tab rename doesn't survive
+// anyway. tmux is no better: #{pane_title} is the same OSC signal, and the one
+// option that records a real rename can't be read without a subprocess per
+// window (see TmuxIntegration).
+//
+// So it can't be trusted as *intent*, but plenty of people title their tabs
+// deliberately and want that name back. `allowTabTitle` is that choice, wired to
+// the "Name sessions from tab titles" setting and false everywhere by default.
```
StuBehan: row label is "Name from tab titles" (Settings.swift:79) and the README matches, this says "Name sessions from tab titles"

StuBehan (changes requested):
Solid, and the rework since I started looking is better than what I was going to suggest. Asking tmux for `#{host}` instead of guessing at `ProcessInfo.hostName` is the right call - my box seeds panes `office_dell_dock.lan` off the dock's DHCP name while `ProcessInfo.hostName` says `StackOne.local`, so the old filter would never have fired here even case-folded. `socketKey` returning nil rather than assuming the default socket is the same instinct and I'm glad its in - `TMUX= cmd` leaving TMUX_PANE behind is a real idiom.

Poked at the other claims and they hold: tmux collapses a literal tab out of `#{pane_title}` and rejects a `select-pane -T` with a newline, `build | tee log` comes through intact, `#{automatic_rename}` renders empty on a never-renamed window, and the host sentinel matched exactly on a fresh pane. 651 tests / 0 failures locally.

One thing left that I'd want fixed - with the toggle on, Claude's `✳` gets read out loud as "eight spoked asterisk". Inline.

The Slack one is a question rather than a change request.

Small stuff: `STACKNUDGE_TAB_TITLE_NAMES` isn't in `notify.conf.example` and the comparable panel toggles are (`STACKNUDGE_MUTE_WHEN_FOCUSED`, `STACKNUDGE_MUTE_DURATION_MIN`, `STACKNUDGE_THEME`). Can you bin off the "Generated with Claude Code" line from the description too, we don't put those in here.

hiskudin:
All four addressed — pushed in d3f2172. `656 tests, 1728 assertions, 0 failures`.

### The `✳` — you were right, and it's worse than pronunciation

I verified both halves of your trace: `expandForSpeech` splits on spaces so the glyph survives as its own word, and the reminder leg (`Panel.swift:2989`, `"Still waiting on \(label)"`) never calls `expandForSpeech` at all. So there was no route to speech that saved us.

But the bigger problem is one neither of us said out loud: **that glyph is an animation frame.** Claude cycles ✳ → ✻ → ✽ as it works, so adopting the title verbatim meant the session's *name* changed on every spinner tick — churning in rows, in banners, and in the event↔session join. The speech was the symptom that made it visible.

So it's trimmed at adoption rather than only for speech, which fixes every surface at once.

One thing that only turned up because I wrote the test as "every frame must reduce to the same name" rather than "the glyph is gone": trimming `CharacterSet.symbols` **isn't enough**. The spinner spans two Unicode categories — ✳ ✻ ✽ are `So` and ∗ is `Sm`, but the middle dot is `Po`. A category-only rule still let the name flip between `Fixing the parser` and `· Fixing the parser`. The stragglers are now listed explicitly rather than widening to all punctuation, which would have eaten the leading bracket of a title typed as `(wip) deploy` (there's a test pinning that too).

Trimmed at the ends only, so `build | tee log.txt → deploy` survives intact. The cost is a deliberate leading emoji — `🚀 deploy` reads as `deploy` — which seems worth paying to stop an animation frame becoming a name.

### Slack — gated, not just documented

You're right that this is a different door. `SlackDelivery.text`'s own comment names `slackIncludeDetail` as the gate for local text on "the one path that leaves the machine", and a tab title written by a `precmd` hook is exactly that class of text. Nothing about `label` used to be program-written; now it is.

So the DM now requires `STACKNUDGE_SLACK_DETAIL` before it will use a tab title, and falls back to the project name otherwise — i.e. someone who turned detail off deliberately gets exactly what they got before this setting existed. On-screen labels and speech are unchanged, since those don't leave the machine. Documented in the README as its own paragraph rather than a parenthetical, because it's a surprising asymmetry.

### Small stuff

`STACKNUDGE_TAB_TITLE_NAMES` added to `notify.conf.example` next to the comparable toggles, comment reworded to match the row label, and the attribution line is off the description. Sorry about that one — noted for future PRs here.

---

Also worth flagging since it changed the diff you reviewed: I ran mutation testing over the new tests and **five mutations left the suite green** — deleting `#{host}` from the format string, changing the delimiter, dropping `-a`, pooling every server's panes into one map, and removing the UTF-8 locale. The format string and `parseTitles` were two halves of one contract with a test on only one half, so deleting a field failed *silently*: no session gets a title again, 649 tests passing. That's fixed with a round-trip test plus `listPanesArgs`/`apply` seams, and I re-verified each mutation is now caught.

Thanks for the `office_dell_dock.lan` data point in particular — that's the exact divergence I could only argue hypothetically from this machine, where the two names differ merely by case.

StuBehan (approved):
LGTM

The spinner catch is better than what I reported - I only traced it as far as speech, but the animation frame churning the name is the actual bug. Checked the category claim too and yeah, `symbols` alone doesn't do it, `·` is Po so it goes straight through.

The five green mutations are the bit thats stuck with me though. A format string and its parser being two halves of one contract, with a test on only one half - I'd bet thats true elsewhere in here, notify.sh especially.

656/0 locally. And the `office_dell_dock.lan` thing was luck rather than insight - my dock hands out a DHCP name, so the two only diverged because I happened to be plugged into it.

PR #170, opened 2026-09-08
fix(voice): recover from a wedged stackvox daemon

StuBehan:
## Problem

Voice notifications go silent with no error once the stackvox TTS daemon wedges. The daemon's worker thread runs `tts.speak()` synchronously; if a `tts.speak()` call hangs on the audio backend (a stale CoreAudio device — headphones yanked mid-utterance, output device swapped), the worker never returns to drain its queue. The queue fills, `submit()` starts returning `busy` (CLI exit 2), and **every** subsequent `stackvox say` is rejected forever.

The app never recovered: `Speaker.speak()` only spawns a fresh `serve` when the socket file is *absent*, but a wedged daemon keeps its socket. And all `say` output was routed to `/dev/null`, so the failure was completely invisible — the fix in the field was a manual `stackvox cancel`/`stop`.

## Fix

- Split `speak()` into a resolver + a private `runSay(…, attempt:)`.
- Added a pure, unit-testable `sayRecovery(exitCode:terminatedBySignal:attempt:)` decision.
- On a clean non-zero `say` exit (busy/wedged), `runSay` sends `cancel()` — which calls `tts.stop()` and unblocks the stuck worker — waits a short beat, and retries once. A signal-terminated `say` (Quit / `stopAllAudio`) always gives up, so we never resurrect audio the user is silencing.
- Added a single `NSLog` on the recovery path so a wedge shows up in Console.app instead of being silent.

Trade-off: if the daemon is legitimately mid-utterance when a new nudge arrives, the retry cancels it to play the newer one (newest-wins) — consistent with `SpeakSelection`'s existing interrupt-on-repeat behavior, and better than dropping the new nudge.

## Testing

New `SpeakerRecoveryTests` (6 cases: clean exit, busy-retry, retry exhaustion, generic non-zero, and both signal-kill cases). Full suite: `make test-without-xcode` → 620 tests, 0 failures.

### DIFF
--- a/Tests/StackNudgePanelCoreTests/SpeakerRecoveryTests.swift
+++ b/Tests/StackNudgePanelCoreTests/SpeakerRecoveryTests.swift
@@ -1,0 +1,46 @@
+import XCTest
+
+@testable import StackNudgePanelCore
+
+final class SpeakerRecoveryTests: XCTestCase {
+
+    // MARK: - sayRecovery
+
+    func test_sayRecovery_acceptsCleanExit() {
+        let actual = Speaker.sayRecovery(exitCode: 0, terminatedBySignal: false, attempt: 1)
+        XCTAssertEqual(actual, .accepted)
+    }
+
+    func test_sayRecovery_retriesOnBusyFirstAttempt() {
+        // CLI exit 2 == daemon replied "busy" (queue full / worker wedged).
+        let actual = Speaker.sayRecovery(exitCode: 2, terminatedBySignal: false, attempt: 1)
+        XCTAssertEqual(actual, .retry)
+    }
+
+    func test_sayRecovery_givesUpAfterFinalAttempt() {
+        let actual = Speaker.sayRecovery(
+            exitCode: 2, terminatedBySignal: false, attempt: Speaker.maxSayAttempts)
+        XCTAssertEqual(actual, .giveUp)
+    }
+
+    func test_sayRecovery_retriesOnAnyNonZeroExit() {
+        // A non-busy failure (e.g. transient client error) is still worth one
+        // cancel+retry — cancel is a harmless no-op when the daemon is down.
+        let actual = Speaker.sayRecovery(exitCode: 1, terminatedBySignal: false, attempt: 1)
+        XCTAssertEqual(actual, .retry)
+    }
+
+    func test_sayRecovery_givesUpWhenKilledBySignal() {
+        // Quit / stopAllAudio SIGTERMs the client — do not resurrect audio the
+        // user is silencing, even though the exit status is non-zero.
+        let actual = Speaker.sayRecovery(exitCode: 15, terminatedBySignal: true, attempt: 1)
+        XCTAssertEqual(actual, .giveUp)
+    }
+
+    func test_sayRecovery_signalKillWinsOverCleanExitCode() {
+        // Defensive: a signal-terminated process should never be read as
+        // accepted, whatever its reported exit code happens to be.
+        let actual = Speaker.sayRecovery(exitCode: 0, terminatedBySignal: true, attempt: 1)
+        XCTAssertEqual(actual, .giveUp)
+    }
+}
--- a/Tests/StackNudgePanelCoreTests/SpeakerRecoveryTests.swift
+++ b/Tests/StackNudgePanelCoreTests/SpeakerRecoveryTests.swift
@@ -23,11 +23,12 @@ final class SpeakerRecoveryTests: XCTestCase {
         XCTAssertEqual(actual, .giveUp)
     }
 
-    func test_sayRecovery_retriesOnAnyNonZeroExit() {
-        // A non-busy failure (e.g. transient client error) is still worth one
-        // cancel+retry — cancel is a harmless no-op when the daemon is down.
+    func test_sayRecovery_givesUpOnDeterministicClientError() {
+        // Exit 1 == empty text or a failed --normalize: deterministic, so a
+        // retry fails identically. Recovering would also cancel any utterance
+        // currently playing for nothing, so only exit 2 (daemon "busy") retries.
         let actual = Speaker.sayRecovery(exitCode: 1, terminatedBySignal: false, attempt: 1)
-        XCTAssertEqual(actual, .retry)
+        XCTAssertEqual(actual, .giveUp)
     }
 
     func test_sayRecovery_givesUpWhenKilledBySignal() {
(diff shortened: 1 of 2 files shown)


hiskudin:
Good catch on the underlying failure — a wedged daemon that keeps its socket is exactly the case `speak()`'s existing "spawn if the socket is absent" check can't see, and routing `say` output to `/dev/null` is why it was invisible. Splitting the policy into a pure `sayRecovery` is the right shape, and unlocking `audioLock` before calling `cancel()` rather than keeping the `defer` is a deliberate, correct detail.

I checked the two assumptions the fix rests on against the installed stackvox. One holds, one doesn't.

**Holds:** signal detection is right. The `SIGTERM`/`SIGINT` handlers are installed inside `serve()` (`daemon.py:317-322`) — the daemon, not the `say` client — so a terminated client really does die by uncaught signal and `terminationReason == .uncaughtSignal` is reliable. Worth knowing it's a dependency on stackvox's internals, but it's currently correct.

---

### 1. The retry outlives Mute and Quit  (medium-high)

`Speaker.swift`, the `.retry` branch.

The guard's stated intent is right — *"we must not resurrect audio the user is actively silencing"* — but `terminatedBySignal` only covers the case where **this** process was killed. It can't cover the case where the process already exited on its own:

1. `say` exits 2 on its own. Clean exit, no signal.
2. The handler removes it from `activeAudio`, then schedules `runSay(attempt: 2)` 200 ms out via `asyncAfter`.
3. Inside that window the user presses `M` or quits.
4. `stopAllAudio()` snapshots `activeAudio` and terminates what's running — but our process is already gone and removed, and **the pending `asyncAfter` isn't tracked anywhere**.
5. 200 ms later a fresh `stackvox say` spawns and speaks.

For Quit it's sharper: `applicationWillTerminate` (`Panel.swift:3279`) exists specifically so quitting silences audio — its comment calls that "the original bug that motivated moving the bell from notify.sh into the app" — and this can spawn a child process after it has run.

The window is small and needs a wedge to open it, but it contradicts the PR's own safety property, so it seems worth closing rather than documenting.

A generation counter is probably the cheapest fix, and unlike the current guard it's testable without spawning anything:

```swift
// bumped by stopAllAudio(), captured at schedule time, checked at fire time
private static var audioGeneration = 0   // guarded by audioLock
```

### 2. `cancel()` fires on deterministic client-side errors  (medium)

`sayRecovery` returns `.retry` for any non-zero exit, and `test_sayRecovery_retriesOnAnyNonZeroExit` encodes that as intended, reasoning that a non-busy failure is *"transient"* and that cancel is *"a harmless no-op when the daemon is down."*

Both halves look wrong against `cli.py`'s `_cmd_say`:

- `return 1` — empty text, or `--normalize` failed (`_resolve_normalized` returned `None`). Deterministic, not transient: the retry re-runs the identical command and fails identically.
- `return 2` — the daemon refused. This is the recoverable case, and it's the only one.

And cancel is a no-op only when the daemon is *down*; when it's **up** it aborts the current utterance and drops the queue. So the case that actually matters is the damaging one.

Concretely: a nudge whose text fails `--normalize` arrives while a previous nudge is mid-sentence. It exits 1, `cancel()` cuts off the sentence the user was listening to, the retry exits 1 again, and it gives up. The user loses audio they should have heard and gains nothing. Before this PR that nudge just failed silently on its own.

Since the codes are cleanly distinguishable, gating on `exitCode == 2` would keep the recovery and drop the collateral — and it matches what the header comment already says the mechanism is for.

### 3. Coverage stops at the pure function  (low)

The six cases all exercise `sayRecovery` in isolation. Nothing covers the recovery *path*: that `cancel()` is actually called, that `attempt` increments, that the retry is scheduled — or the mute/quit interaction in #1, which is where the real defect is. Understandable while `runSay` is private and spawns processes, but the generation-counter fix in #1 would be directly unit-testable, so that one at least could land with a test.

### 4. Nit

`NSLog` treats its first argument as a format string. Only an `Int32` is interpolated today so it's safe, but it's a latent hazard if `text` is ever added to that line. `NSLog("%@", "...")` sidesteps it permanently.

---

#1 and #2 both look worth resolving before merge; #3 and #4 are yours to weigh. Happy to be wrong on any of this — particularly #2, if exit 1 turns out to be reachable in a way that genuinely benefits from the retry.

StuBehan:
@hiskudin good spot on #1, that ones a proper hole - the retry was sat in an asyncAfter that stopAllAudio never sees, so if you mute or quit in that 200ms gap it speaks anyway. gone with the generation counter - bumped in stopAllAudio, captured when the retry is scheduled, checked at fire time via retryStillValid (so its actually unit tested now). both mute and quit route through stopAllAudio so it covers both.

#2 yeah I had this backwards - only exit 2 (busy) retries now. exit 1 is deterministic so the retry just fails the same way, and cancelling a playing utterance to do it was the damaging bit. lines up with what the header comment claimed the mechanism was for anyway 😅

#4 done (NSLog("%@", ...)). #3 - the generation check gives me a second pure fn so theres a test for that, left the cancel/attempt++ path as is since its still spawning processes.

thanks, solid review

hiskudin:
Pulled the branch and checked all four — `./build.sh` clean, `make test-without-xcode` → **622 tests, 1654 assertions, 0 failures**.

- **#1** — generation counter is correct and `&+=` avoids an overflow trap. Every access (`stopAllAudio`, the handler capture, the fire-time read) is under `audioLock`, and I confirmed mute and quit are the only two `stopAllAudio()` callers (`Panel.swift:2650`, `:3279`), so it does cover both.
- **#2** — `exitCode == 2 ? .retry : .giveUp` matches `_cmd_say` exactly.
- **#4** — `NSLog("%@", …)` ✅.

One residual on #1, and it's genuinely low — the window went from 200 ms to sub-millisecond, so this is a "while you're here", not a re-request.

The generation is captured *inside* the termination handler (`Speaker.swift:109`), so there's still a small ordering gap: if Mute lands after `say` exits but before the handler runs, the handler reads the **already-bumped** generation, the fire-time comparison matches, and the retry speaks anyway. Foundation dispatches `terminationHandler` asynchronously, so that gap is real but tiny, and under load it's a few ms rather than zero.

Moving the capture to `runSay` entry — before `say.terminationHandler` is assigned — closes it, since any bump after the utterance was requested then invalidates it. Same one comparison, just read earlier. It also reads more like the property you actually want: *don't play audio that was requested before the user asked for silence*.

Quit is mostly moot either way (the process is going away and the `asyncAfter` won't fire), so Mute is the only case this affects.

Nice fix on #2 in particular — dropping the cancel on deterministic failures is the part that was quietly costing users audio they should have heard.

hiskudin (approved):

PR #173, opened 2026-09-11
fix(notify): stop nudging about answered prompts, and rate-limit finished-turn DMs

hiskudin:
Two notification bugs found in daily use, both diagnosed from live evidence on the reporting machine rather than from reading the code.

---

## 1. Answering a prompt elsewhere didn't clear it

**Symptom:** approve a plan in the terminal and the notification stays — in the panel *and* in Slack — and keeps re-nudging.

**Cause:** the panel treated *"the FIFO file exists"* as proof a prompt was still blocking. That holds only while the hook reading it is alive, and `notify.sh` cleans up with a `trap … EXIT` — which bash honours for `SIGTERM` (I tested this) but nothing honours for `SIGKILL`, which is what the agent does to the hook when you answer in its own UI. The file outlives the process, so the panel kept the prompt in the menu-bar count and fired reminders at the user and at Slack for the full 550s.

**Evidence:** 536 leaked FIFO directories in `$TMPDIR` going back three months, **every one still holding a live FIFO** — so the trap had not run once. Newest was from the previous day.

**Fix:** the hook's pid now travels with the event. The FIFO answers *"was a prompt raised?"*; the pid answers *"is anyone still listening?"*. Both are required, so a prompt whose hook has gone retires within a tick. An event from an older `notify.sh` carries no pid and falls back to the previous behaviour, rather than having Allow/Deny silently disabled for the first event after an upgrade.

Also traps `INT/TERM/HUP` explicitly and sweeps leaked dirs older than 30 minutes — comfortably past the 550s a prompt can live, so one in use is never in range. Neither is load-bearing now the panel doesn't trust the trap, but 536 leaked FIFOs is its own small problem.

> **Considered and rejected:** probing with `open(O_WRONLY|O_NONBLOCK)` and reading `ENXIO` as "no reader". Needs no new field — but the hook `select`s on the read end, so a probe opening and closing the write end delivers EOF, the hook exits with an empty decision, and the panel's Allow/Deny stops working. Worse than the bug being fixed.

## 2. An hour away was fifty Slack DMs

**Cause:** the idle threshold is a **floor, not a rate limit**. Once you cross it the condition stays true for the whole absence, so every event passed it independently. The setting answers *"are you away?"* and nothing answered *"how often should I interrupt you while you are?"*

**Evidence:** the reporting machine's event log — 1284 records, **96% of them `stop`**, with busy hours running 50–56 events. With "also notify on finished turns" on, an hour away was ~50 DMs, not the six a ten-minute idle setting implies.

**Fix:** finished turns collapse into one message per 15 minutes, and that message says what it swallowed (*"finished a turn · 7 more while you were away"*) so a quiet hour reads as one message about eight turns rather than looking like eight went missing. The cooldown resets while the user is at the machine, so the first stop of each absence still arrives promptly. With the idle gate on `Always` there's no "present" to detect and it runs continuously, which is what that setting asks for.

**Permission prompts are deliberately exempt** — each blocks an agent until answered, they're rare (45 of those 1284), and repeats of a single one are already capped by `maxReminders`. Throttling them would withhold precisely the notifications that are actionable.

---

## Testing

`./build.sh` clean · `make test-without-xcode` → **678 tests, 1773 assertions, 0 failures**

New coverage is policy-level and injectable (`AttentionPolicy.isAnswerable`, `SlackDelivery.throttleStop`), including:

- a FIFO that outlived its hook is not answerable; a missing pid falls back to the old behaviour
- liveness isn't consulted for events that can never qualify
- the burst test asserts **conservation** — every finished turn is either reported in a sent message or still pending in the suppressed count. A message count alone passed while silently dropping events, which is the failure that actually matters.

The `find` sweep was dry-run against the real 536 directories: all matches our own pattern, nothing else caught.

## Note for whoever merges

`feat/slack-permission-responses` also modifies `notifySlack`, so whichever lands second needs a small manual resolution.


hiskudin:
Ran an adversarial pass over this before asking anyone else to look. It found one real bug, now fixed in `0adcd12`.

### Fixed — the throttle's headline property was false end to end

`resetStopThrottleIfPresent` zeroed `suppressedStopCount` along with the cooldown, so turns that finished while you were away and hadn't been reported yet were **discarded** rather than folded into the next message.

Easy to hit without ever really returning: idle is measured from the last HID event, so a stray trackpad bump or a notification click reads as "present", the pending count is wiped, and the next DM claims a bare *"finished a turn"* as though nothing preceded it.

The burst test asserted conservation over `throttleStop` **alone**, which is exactly why it didn't catch this — the loss happened in the controller, outside the function under test. A regression test now covers the return-to-machine path.

Only the cooldown is cleared now. Keeping the count is accurate rather than merely safe: it's incremented only past the idle gate in `notifySlack`, so it can only ever hold turns that genuinely finished while you were away, however many brief presences that absence gets split across.

### Checked and dismissed

- **Pid reuse making a dead prompt look alive** — bounded and benign, and I've added the reasoning to the code since it's the obvious objection. A watch is created with `firstSeenAt` taken from the event and retired in the same pass once that exceeds `promptLifetime`, so a reused pid can't resurrect a stale prompt. The window is 550s from the prompt, *not* the 30-minute sweep interval, and inside it the worst case is simply the behaviour this PR replaces.
- **Data race on the throttle counters** — none. `EventStore.append` always arrives via `DispatchQueue.main.async`, and the ticker is a main-runloop `Timer`, so everything touching that state is single-threaded.
- **Stale pid surviving a restart** — `EventLog` deliberately excludes `fifoPath`/`hookPID` from the persisted record ("a replayed record cannot masquerade as actionable"), so a reloaded event never carries one.
- **Sweep safety** — `find -type d` doesn't match symlinks and `rm -rf` doesn't dereference them; cross-user removal would need `$TMPDIR` unset *and* would still be blocked by the sticky bit on `/tmp`.
- **Zombie processes answering `kill(0)`** — real but self-corrects on the next 5s tick.
- **Double cleanup from `trap … EXIT INT TERM HUP`** — happens, harmless: `rm -f` / `rmdir 2>/dev/null` are idempotent.

### Known gap, not closed here

The real `kill(pid, 0)` wiring and all of `notify.sh` are untested — `bash -n` is the only check the shell script gets. I verified the liveness path empirically instead, reproducing the exact bug against a faithful mock of the hook's structure:

```
recorded $$ = 14023
is that pid the live blocking script?   bash          <- $$ is correct
kill(pid,0) while alive:                ALIVE
--- SIGKILL, as the agent does ---
FIFO still exists after SIGKILL:        YES  (the leak)
kill(pid,0) after:                      DEAD -> prompt retires
```

That's evidence, not coverage. A bash test harness for `notify.sh` is worth its own PR — it'd also cover the `stop`/`permission` payload shape, which is the same class of untested producer/consumer contract that let a format string break silently in #169.

`./build.sh` clean · **679 tests, 1775 assertions, 0 failures**

notify.sh
```diff
@@ -732,7 +738,19 @@ notify_macos() {
...
+# live for, so a dir in use is never in range — and errors ignored, because a
+# sweep that fails must never delay the notification it runs alongside.
+sweep_stale_perm_fifos() {
+  local root="${TMPDIR:-/tmp}"
+  find "$root" -maxdepth 1 -name 'stack-nudge-perm.*' -type d -mmin +30 \
+    -exec rm -rf {} + 2>/dev/null || true
+}
+
 create_perm_fifo() {
+  sweep_stale_perm_fifos
```
StuBehan: this runs before `post_to_panel` though right? so its on the path to the banner, not alongside it.

my `$TMPDIR` has 30,791 entries in it and `-maxdepth 1` still stats every one of them for `-type d` - first run took 2.2s, warm it's 0.05s. So on a cold cache that's two seconds before I find out an agent is blocked, on the one path where latency is basically the product!

can we not just background it? the dir you've just made is seconds old so it can never be in range of `-mmin +30`, and the hook hangs around for 550s anyway so the find has plenty of time.

StuBehan (commented):
Solid pair of fixes and I think both diagnoses are right - the pid and the fifo answering two different questions is the bit that makes it work, rather than trying to make the trap more reliable.

Pulled it and ran `make test-without-xcode`, 679 tests / 1775 assertions / 0 failures, same as you got.

Two things I'd want sorting before this merges - the sweep sitting on the hot path, and the suppressed count outliving the absence it belongs to. Rest is questions and nits, flagged as such.

Also on the zombie point in your adversarial pass - it self corrects because node reaps its children more or less straight away, but that's the load bearing bit isn't it? Worth a clause in the comment, a SIGKILLed child still answering `kill(0)` is exactly the case this whole thing exists for.

panel/Panel.swift
```diff
@@ -2915,9 +2941,33 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+        // not been reported yet, and zeroing it here threw them away. Idle is
+        // measured from the last HID event, so a stray trackpad bump — or a
+        // notification click — counts as "present" and would silently discard
+        // the record of everything missed, leaving the next DM claiming a bare
+        // "finished a turn" as though nothing had preceded it.
+        //
+        // Carrying it is accurate rather than merely safe: the counter is only
+        // ever incremented past the idle gate in notifySlack, so it can only
+        // contain turns that genuinely finished while the user was away.
+        lastStopSlackAt = nil
```
StuBehan: "can only contain turns that genuinely finished while the user was away" - agreed, but is it always *this* absence?

ie I'm away, 4 DMs go out, last 10 turns get swallowed. I come back and work all afternoon, and the idle gate means nothing increments the count but nothing clears it either. Then I walk off tomorrow morning, one turn finishes, and the first DM says "10 more while you were away" about yesterday.

I think the discriminator you want is a present streak longer than the idle threshold - one trackpad bump keeps idle under the threshold for exactly `slackIdleMinutes` and not a second more, so anything past that means two HID events close together, ie an actual human. That'd keep the bump fix in `0adcd12` and bin off the stale carry over 🤔

panel/Panel.swift
```diff
@@ -2915,9 +2941,33 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
         if nav.slackTokenPresent { refreshSlackStatus() }
     }
 
+    // Clear the finished-turn cooldown while the user is actually at the machine,
+    // so the first stop after they walk away is sent immediately rather than
+    // being swallowed by a window left over from the previous absence. With the
+    // idle gate set to Always there is no "present" to detect, so the cooldown
+    // runs continuously — which is what that setting asks for.
+    private func resetStopThrottleIfPresent() {
+        guard nav.slackIdleMinutes > 0,
```
StuBehan: `Always` is 0, so this returns early and the cooldown never resets - meaning with the idle gate off stop DMs are capped at one per 15 mins permanently, including while I'm sat right in front of it.

README still says "Set it to `Always` if you'd rather have everything" one bullet above the new one you added. Is that the behaviour we want? Always feels like the one setting where someone has explicitly asked for the firehose.

panel/Panel.swift
```diff
@@ -2915,9 +2941,33 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
...
+        // the record of everything missed, leaving the next DM claiming a bare
+        // "finished a turn" as though nothing had preceded it.
+        //
+        // Carrying it is accurate rather than merely safe: the counter is only
+        // ever incremented past the idle gate in notifySlack, so it can only
+        // contain turns that genuinely finished while the user was away.
+        lastStopSlackAt = nil
+    }
+
     // A permission prompt we can *prove* is still blocking: notify.sh creates
```
StuBehan: this comment is the belief the PR disproves - "so the file existing means nobody has answered yet". Worth updating while you're in here, its directly contradicted by the new one in `isAnswerable`. Same for "Retire on the FIFO alone" at 3006, its fifo *and* liveness now.

panel/Panel.swift
```diff
@@ -2928,8 +2978,16 @@ final class PanelController: NSObject, NSApplicationDelegate, PanelKeyDelegate,
     // reminder we can't verify would eventually fire for prompts already
     // handled in the terminal, and a nudge that cries wolf is worse than none.
     private func isBlockingPrompt(_ event: NudgeEvent) -> Bool {
```
StuBehan: once this goes false the watch retires and the menu bar count clears, but the row stays in the panel with its Allow/Deny, and `writeFIFO` will get ENXIO and return silently. So I can still click Allow on a prompt we now *know* is dead and nothing happens at all?

not saying the row has to vanish, queueing is how the rest of the list behaves - but now that we can actually tell the difference, should the buttons go?

panel/SlackNotifier.swift
```diff
@@ -77,7 +113,11 @@ enum SlackDelivery {
                 ? "\(subject) is still waiting for permission"
                 : "\(subject) needs permission"
         case .stop:
-            headline = "\(subject) finished a turn"
+            // Say what was folded in, so a quiet hour reads as one message about
+            // eight turns rather than looking like eight turns went missing.
+            headline = coalesced > 0
+                ? "\(subject) finished a turn · \(coalesced) more while you were away"
```
StuBehan, line 150: these counters live on the controller so they're global across sessions right? with 3 agents running the headline names whichever repo won the race and the other two collapse into an anonymous number, so "stack-nudge finished a turn · 7 more" where the 7 were somewhere else entirely.

Is per session throttling worth it, or are we happy the DM just means "go look at your machine"? (happy either way tbh, just don't think its written down anywhere as a choice)

panel/EventListener.swift
```diff
@@ -213,6 +214,7 @@ private struct NudgeEventDTO: Decodable {
             sessionID: session_id,
             itermTabName: iterm_tab_name,
             fifoPath: Self.validatedFifoPath(fifo_path),
+            hookPID: hook_pid,
```
StuBehan: `fifo_path` gets `validatedFifoPath` but the pid goes straight through - a 0 or a negative means `kill(0, 0)` / `kill(-1, 0)`, both of which return 0, so every prompt reads alive. Degrades to the old behaviour rather than anything worse so not gonna block for it, but feels odd to harden one and not the other when they arrive on the same socket.

Tests/StackNudgePanelCoreTests/SlackDeliveryTests.swift
```diff
@@ -133,4 +133,120 @@ final class SlackDeliveryTests: XCTestCase {
...
+
+        // They leave again and another turn finishes: it sends at once, and
+        // reports the three that were waiting.
+        XCTAssertEqual(SlackDelivery.throttleStop(now: t0.addingTimeInterval(60),
+                                                  lastSentAt: lastSent,
+                                                  suppressed: suppressed),
+                       .send(coalesced: 3),
+                       "turns missed while away must not be dropped on a stray keystroke")
+        suppressed = 0
+        XCTAssertEqual(suppressed, 0)
```
StuBehan: this is asserting on a local you assigned on the line above so it can't fail 🤷 the real assertion is the `.send(coalesced: 3)` above it.

panel/SlackNotifier.swift
```diff
@@ -58,14 +58,50 @@ enum SlackDelivery {
...
+    // A stop DM says "your agent finished, come back". Once you know that, the
+    // next seven say nothing new, so they are folded into a count on the next
+    // one that gets through rather than dropped silently.
+    //
+    // Permission prompts are deliberately exempt. Each blocks an agent until it
+    // is answered, they are rare (45 of those 1284), and repeats of any single
+    // one are already bounded by AttentionPolicy.maxReminders. Throttling them
+    // would withhold the notifications that are actually actionable — the more
+    // so now that a prompt can be answered from Slack.
+    static let stopCooldown: TimeInterval = 15 * 60
```
StuBehan, line 78: any appetite for making this cyclable like the idle threshold sat right next to it? feels like the number most likely to wind someone up, 15 mins is a long wait if your turns are short. Not for this PR necessarily.

notify.sh
hiskudin: You're right on both counts — it ran inside `create_perm_fifo`, which is called before `post_to_panel`, so it was squarely on the path to the banner. 2.2s cold is damning; I'd reasoned "it's just a find" and never measured it on a realistic $TMPDIR.

Backgrounded in `1bc3f47`. The reasoning you gave is exactly why it's safe, so I've put it in the comment: the dir we've just created is seconds old and can't be in range of `-mmin +30`, and the hook sticks around 550s, so the sweep has all the time it wants.

panel/Panel.swift
hiskudin: Yes — "while you were away" was doing work the code couldn't back, and your scenario is the one that breaks it. I'd traded a dropped count for a stale one and only checked the first.

Taken your discriminator verbatim, it's better than what I'd have reached for. A lone HID event keeps idle under the threshold for exactly the threshold and not a second longer, so it can never present as sustained; two events far enough apart can, and that's a person. Three-way now — `away` / `brieflyPresent` (cooldown only) / `sustainedPresent` (clears the backlog too).

Tested by walking the whole life of a single bump in 30s steps and asserting it stays `brief` throughout, rather than spot-checking one value either side.

panel/Panel.swift
hiskudin: Good catch that the README contradicted the code one bullet apart.

I've fixed the docs rather than the behaviour, and I'd argue that's the right way round: `Always` asks to remove the *idle* gate — send even though I'm at my desk — and shipping a path that still does fifty DMs an hour would leave the exact defect this PR exists to fix, just behind one setting. README now says so explicitly.

Genuinely open to being wrong if you think Always should mean no limit at all. It'd be a `cooldown: 0` special case, but say the word — I'd rather that than have you merge a behaviour you disagree with.

panel/Panel.swift
hiskudin: Fair — a comment asserting the thing the PR disproves is worse than no comment. Both updated in `1bc3f47`: the first now says the file is necessary but not sufficient and why, the second reads "the FIFO *and* its hook's liveness".

panel/Panel.swift
hiskudin: No, that's a real hole and I'd missed it — the count clears but the row keeps a button that silently does nothing, since `writeFIFO` gets ENXIO and returns.

Buttons now go with the liveness. I kept the row (queueing is how the rest of the list behaves, as you say) and only dropped the affordance. The check moved to `NudgeEvent.isStillBlocking` so the view layer and the controller ask the same question the same way, instead of me duplicating the wiring.

panel/SlackNotifier.swift
hiskudin: Correct, they're on the controller, so it's global across sessions. That was a decision I'd made implicitly and never wrote down, which is the fair part of the complaint.

Keeping it global: per-session would mean N DMs per window for N agents, which is most of the way back to the bug. The DM's job is "go and look at your machine", not "look here specifically". Now stated in the README so the next person meets it as a choice rather than inferring it from behaviour.

panel/EventListener.swift
hiskudin: Agreed on the inconsistency, and it's worse than you flagged — I checked both against the real syscall.

`kill(0, 0)` signals the caller's whole process group and `kill(-1, 0)` every process it may signal; both return 0, so either value makes *every* prompt read alive and quietly undoes the fix. And `pid_t` is Int32, so `pid_t(4_000_000_000)` **traps** — an oversized number on that socket crashes the panel outright, which isn't a graceful degrade at all.

Validated in `1bc3f47` (`AttentionPolicy.validHookPID`, tested), and the conversion itself made non-trapping rather than left one guard away.

Tests/StackNudgePanelCoreTests/SlackDeliveryTests.swift
hiskudin: Guilty, and on a PR arguing for conservation over counting. Removed — the `.send(coalesced: 3)` above it is the assertion that means anything.

panel/SlackNotifier.swift
hiskudin: Agreed it's the number most likely to annoy, and cycling it next to the idle threshold is the obvious home. Leaving it out of this PR — it needs a row, a config key and an options list, and this one is already carrying two fixes and nine review threads.

Left open deliberately so it doesn't get lost; happy to do it as a follow-up if you want it before the next release.

hiskudin:
All nine addressed in `1bc3f47` — replies on each thread, seven resolved, two left open on purpose (the `Always` question and the cyclable cooldown, both of which are yours to call).

**686 tests, 1806 assertions, 0 failures** · `./build.sh` clean.

Both blockers were right, and the sweep one I should have caught myself — I reasoned "it's only a `find`" and never measured it on a realistic `$TMPDIR`. 2.2s in front of the banner is the worst possible place to put it.

The two I'd single out:

**The presence discriminator is your design, not mine.** I'd fixed "a bump discards the backlog" and created "yesterday's backlog reports tomorrow" without noticing I'd only moved the problem. Using the *duration* of the presence separates them cleanly, and it's testable in a way my version wasn't — the test now walks the entire life of a single bump and asserts it never once looks sustained, rather than spot-checking either side of a boundary.

**The pid finding was worse than flagged.** You called it "degrades to the old behaviour, not blocking". It doesn't: `pid_t` is Int32, so `pid_t(4_000_000_000)` **traps**, and that value arrives on a local socket. That's a crash, not a degrade. Validated, and the conversion made non-trapping too rather than relying on the one guard.

Also fixed the two comments that asserted the belief this PR disproves — that's the kind of thing that outlives the code and misleads whoever reads it next.

On the zombie point from your summary: agreed, and it's in the comment now. It self-corrects *because* the agent reaps promptly, and that's load-bearing rather than incidental — a SIGKILLed child still answering `kill(0)` is exactly the case this check exists for, so an agent that left hooks unreaped would quietly degrade this back to FIFO-only. Never worse than before, but no longer a fix.

PR #175, opened 2026-09-14
feat(panel): per-client usage errors and reset clock times

StuBehan:
Two things, both in the Usage tab.

The usage was flakey and the error state only ever said Claude. The `claude /usage` shell-out is the one probe that hits the network, so it fails a fair bit more than the local Codex/Antigravity reads - and on a fail it was nulling the snapshot, so the whole Claude row dropped out of the sidebar and flickered back on the next tick. Now it holds the last-good snapshot and marks it stale, and the error is per-client so the pane names whatever actually failed rather than always pointing at Claude. A missing `claude` on PATH is silent too now, so a Codex-only user doesn't get a phantom Claude error.

Second, reset times. Codex reports its reset as a unix timestamp and Antigravity as ISO 8601, so they all render in the same clock format Claude's CLI prints ("in 2 hours · Jun 30 at 6:50pm"), in the Usage tab and the quota banner.

Shared `QuotaSnapshot.hasTier` falls out of both - a snapshot can parse ok and still hold no tier we map, and the two callers (which clients get a row, and whether an error replaces the bars or sits above them) have to agree on that.

683 tests, all green.


hiskudin (commented):
Read through this properly — the shape is good and two bits are better than they needed to be. `hasTier` is the right call: "non-nil snapshot" and "has anything to draw" genuinely are different questions and having two callers answer them differently was a real bug. And `testAbsoluteLabelRoundTripsThroughTheClaudeParser` is exactly the producer/parser pinning this repo keeps needing — render it, push it back through `parseResetsAt`, assert the same instant. That's the class of test the #169 mutation run showed we were missing.

Three things worth a look, one of which is mine to deal with.

## 1. Holding a stale snapshot collides with the pace bars in #176

This is the one I'd want settled before both land, and it's my problem more than yours.

Hard-fail used to null `nav.quota`. This PR holds the last-good snapshot and marks it stale, which is the better behaviour on its own terms. But #176 adds an elapsed-time bar to each tier row, derived from `resetsAt` and recomputed against `Date()` on every redraw — while `utilization` stays frozen at whatever the held snapshot says.

So under a sustained `claude` failure the pale elapsed bar keeps advancing while the solid usage bar sits still, and the row reads as "increasingly comfortably under pace" from data that's minutes or hours old. Nulling the snapshot used to make that impossible; after this it's reachable. The stale note helps, but the bar is the thing people actually read.

Not asking you to change anything — `quotaErrors[client]` is exactly the signal needed, and you've deliberately left `quotaClaudeLastUpdated` untouched so the age is still available. I'll gate the elapsed bar on it in #176. Flagging it so the interaction is on the record rather than discovered after both merge.

Also worth knowing: these two PRs touch the same five production files and both rewrite the `Resets …` line in `tierRow`, so whichever lands second will conflict there. Small, but not zero.

## 2. `QuotaReset.absoluteLabel` mutates a shared formatter

`onTheHourFormatter` / `withMinutesFormatter` are cached statics, and `absoluteLabel` does `formatter.timeZone = timeZone` on the way through. The comment says this is reused "main thread only, as with RelativeTime" — but `RelativeTime` only ever *reads* its cached formatters, so the parallel doesn't quite hold. This one writes.

Every current caller is main-thread (`tierRow` in a view body, `postQuotaBanner` off the probe completion), so it's latent rather than live. But `DateFormatter` mutation isn't thread-safe, and the next caller won't know the constraint exists. Cheapest fixes in rough order of preference: construct the formatter inside the function, key the cache by timezone, or leave the cache and add a `dispatchPrecondition(condition: .onQueue(.main))` so a future off-main caller trips immediately instead of returning a quietly wrong string.

## 3. `cliMissing` drops a snapshot the rest of the PR would have kept

`ProcessOutput.claude()` returning nil sets `cliMissing`, and the handler then nils `nav.quota` and `quotaClaudeLastUpdated`. The intent is right — a Codex-only user shouldn't get a phantom Claude error — but this ends up being the one path in the PR that discards good data, which cuts against everything else here.

`claude` not resolving on one tick isn't strong evidence of "not a Claude user". PATH under a launched `.app` is materially different from PATH in a shell, and this app's launch environment is already known to be odd (`make reload`'s `launchctl kickstart` doesn't find the service at all on my machine). A single unlucky resolution silently removes Claude from the sidebar entirely.

Suggestion that keeps your intent: treat `cliMissing` as "not a Claude user" only when there's no prior snapshot. If one exists, hold it and mark it stale like every other failure here. Someone who genuinely doesn't have Claude never had a snapshot, so they still get silence.

## Minor

`absoluteLabel` renders in the machine's timezone while Claude's CLI prints the account's, so the Usage tab and `claude /usage` will show different clock times for anyone whose account timezone isn't their machine's. Your comment calls this out and the reasoning holds — the countdown is measured against local — so this is a note, not a request. Might be worth a word in the README so it reads as deliberate rather than a bug.

hiskudin (approved):

PR #176, opened 2026-09-14
fix(panel): show a pace marker on Usage tab quota bars

hiskudin:
Requested by Kenneth Ng: a way to see time-elapsed against quota in the Usage tab, "to easily see if current usage is exceeding remaining quota, linearly projected ahead (e.g. am I above 50% token usage with 50% time remaining until refresh)".

## What changed

Each quota bar gains a tick at the point the window has reached. Fill behind it means you'll coast to the reset; fill past it means you'll hit the cap early.

It's an overlay on the existing bar rather than a second bar, for two reasons. The comparison is the feature — two bars at different vertical positions leaves the reader doing the arithmetic. And the quota pane fits only ~3 tier rows at the default 600x320 panel while Claude Max renders 4, so a row that grew would push one off screen. Row pitch is unchanged.

No new setting: the marker is passive and costs no height, so there's little to opt out of.

## The mislabel this uncovered

Positioning needs a window length, which `QuotaTier` didn't carry. Codex publishes one per limit as `window_minutes` and `CodexUsage` was reading past it.

That field turns out to be the only reliable way to tell Codex's two windows apart. The tab titled tiers by slot, assuming `primary` = 5h and `secondary` = weekly:

| `limit_id` | `primary` | `secondary` |
|---|---|---|
| `codex` | 10080m = 7 days | absent |
| `codex_bengalfox` | 300m = 5h | 10080m = 7d |

Usually `primary` carries the weekly window with `secondary` absent — so a weekly quota was displayed under "Current session (5h)", with no weekly row. But the mapping isn't even fixed per `limit_id`: 28 of 11,792 `codex` lines locally carry the 300/10080 pair instead. Neither the slot nor the id can stand in for the window, so headings are now named from `window_minutes`. Verified on screen: that tier now reads "Current week".

Across 60 local rollouts (23,696 non-null limit objects), `window_minutes` was present every single time.

Claude's `/usage` text publishes no window length, so it's inferred from the tier name. Antigravity reports neither a length nor, for monthly credits, a reset time — those get no marker.

## Testing

708 tests green. New: `QuotaPaceTests` (fraction boundaries, clamping, the absent cases) and `CodexUsageTests` (verbatim rollout lines through the real parser).

Mutation testing found a real gap: swapping the Claude weekly tier's window to 5h left the whole suite green, while pegging every weekly marker to the far left forever. That mapping is now pinned by driving the real `/usage` text through the real parser, rather than testing the helper alone — the contract-gap lesson from #169.

Also verified against live data: the Codex tier reads 13% used at 32% elapsed both in the panel and via an independent probe of the rollout file.

## Note for the reviewer

`make reload` does not restart the app on my machine — `launchctl kickstart` fails with "Could not find service com.stackonehq.stack-nudge", and `|| true` swallows it, so it silently reports "reloaded" while the old binary keeps running. Not touched here, but worth a follow-up.


panel/Formatting.swift
```diff
@@ -110,6 +119,63 @@ enum QuotaReset {
...
+
+    // How far through its window a tier is, 0…1. Shares `remaining`'s
+    // past-deadline rule: a stale snapshot draws no marker rather than one
+    // pinned confidently to the far right.
+    static func elapsedFraction(until date: Date,
+                                windowLength: TimeInterval,
+                                now: Date = Date()) -> Double? {
+        guard windowLength > 0, let remaining = remaining(until: date, now: now) else { return nil }
+        // Clamped, not rejected: a window longer than we assume pins to the start.
+        return min(1, max(0, 1 - remaining / windowLength))
```
StuBehan, line 131: Clamping means a window we've guessed wrong draws a confident warning instead of nothing. Codex reports its window so it can't hit this, but Claude's is inferred from the tier name, so this branch is the wrong-inference case. 30% used, reset 9h out, 5h assumed window gives elapsed 0.0 and "30% ahead of pace" - the overshoot is just the utilisation. Your comment at the top of the func already has the rule. Should we not return nil above the window length? testRemainingLongerThanWindowClampsToZero pins the clamp but nothing checks what paceWarning does with it. (not reachable today afaict, parseResetsAt caps at 60 days, so not a fire)

panel/ClaudeCliQuotaProbe.swift
```diff
@@ -208,7 +208,17 @@ final class ClaudeCliQuotaProbe {
...
+        return (name, QuotaTier(utilization: pct,
+                                resetsAt: resetsAt,
+                                windowLength: windowLength(forTier: name)))
+    }
+
+    // The `/usage` text reports a reset time but never a window length, so it
+    // has to come from the tier name.
+    static func windowLength(forTier name: String) -> TimeInterval? {
+        if name == "session" { return QuotaWindow.fiveHours }
+        if name.hasPrefix("week") { return QuotaWindow.sevenDays }
```
StuBehan, line 220: prefix match, so anything named `week (...)` gets 7d whatever its real cadence. Fine today, just noting it's the input to the clamp above.

panel/SessionUsage.swift
```diff
@@ -551,18 +562,112 @@ struct UsageView: View {
...
+                // .orange is already a severity step on the widget's ramp, and
+                // orange text under a red bar reads as less urgent than the bar
+                // it describes.
+                HStack(spacing: 4) {
+                    Text("Resets \(label)").foregroundStyle(.tertiary)
+                    if let warning = Self.paceWarning(tier) {
+                        Image(systemName: "exclamationmark.triangle.fill")
+                            .font(.system(size: 9))
+                            .foregroundStyle(.orange)
+                        Text(warning).foregroundStyle(.secondary)
```
StuBehan, line 588: paceDescription already puts the warning in the bar's accessibilityValue, so VoiceOver reads it twice. Hide one?

panel/SessionUsage.swift
```diff
@@ -551,18 +562,112 @@ struct UsageView: View {
...
     }
 
+    // "22% ahead of pace", or nil when usage isn't meaningfully ahead of the
+    // clock. Reachable from tests; the view around it is not.
+    static func paceWarning(_ tier: QuotaTier, now: Date = Date()) -> String? {
+        guard let elapsed = elapsedFraction(tier, now: now),
+              let overshoot = QuotaReset.paceOvershoot(utilization: tier.utilization,
+                                                       elapsedFraction: elapsed)
+        else { return nil }
+        return "\(Int(overshoot.rounded()))% ahead of pace"
```
StuBehan, line 604: that's percentage points printed with a %, next to a real percentage in the row above. Reads fine to me but flagging it.

StuBehan (commented):
Codex mislabel find is great, and pinning it through the real parser rather than the helper is the right call. Sharing the Resets line so the warning can't reflow the row is nicer than what I was going to suggest.

The PR body is still describing the old design though - "gains a tick", "an overlay on the existing bar rather than a second bar", "the marker is passive and costs no height". README got fixed, body didn't. Worth a pass, that's the bit we read back in six months. The warning is also the biggest new thing on screen and the body doesn't mention it at all.

One inline I think is a real problem, rest is small.

PR #177, opened 2026-09-14
fix(notify): stop nudging about plans already approved in the terminal

hiskudin:
Reported in use: approving a plan in Claude Code's own UI leaves stack-nudge reminding about it — locally and in Slack — for roughly nine more minutes.

## Cause

`ExitPlanMode` took the blocking permission path. `notify.sh` created a FIFO and waited on it for an allow/deny decision, because `is_question_event` only excluded `AskUserQuestion`.

Answering in CC's own UI doesn't end that hook, so it ran to its full 550s timeout. For the whole of that window the panel saw a live hook holding a live FIFO — precisely its definition of "a prompt is still waiting". Reminders are capped at `maxReminders = 3`, so the nagging is bounded at three; what persisted the full 550s was the menu-bar pending count and an Allow/Deny affordance that could no longer do anything.

**The stronger justification is the second one below**, and it doesn't depend on the timing story: an `allow` on ExitPlanMode emits `{"behavior":"allow"}`, approving the plan outright — while the banner body is the bare string `ExitPlanMode`, so no plan text ever reaches the panel or Slack. The button approved something its approver could not see.

#173 fixed the neighbouring case, retiring a prompt whose hook has *died*. This hook hadn't died; it was still blocked on a FIFO nobody was ever going to write to.

## Fix

`ExitPlanMode` belongs with `AskUserQuestion`. Both are answered in the agent's own UI, and for both an "allow" written from the panel means "take the default" — which here means approving a plan the user hasn't read. With `has_action=false` no FIFO is created, panel Enter falls through to focusing the editor, and no reminder loop starts.

## Verification

Against the installed hook, before and after:

| tool | FIFO created | hook blocks |
|---|---|---|
| `ExitPlanMode` (before) | yes | yes, to the 550s timeout |
| `ExitPlanMode` (after) | **no** | **no — returns in 0s** |
| `AskUserQuestion` | no | no (unchanged) |
| `Bash` | yes | yes (unchanged) |

Also confirmed directly that a SIGKILLed hook leaves its FIFO directory behind, which is why FIFO existence alone can't stand in for liveness.

## Accepted trade-off

ExitPlanMode now gets one banner and one Slack DM, and no reminders or menu-bar count — the same deal AskUserQuestion has had. A plan approval is the canonical "agent is blocked until you come back" event, so that is a real cost.

It isn't avoidable at this layer: without a blocking hook there is no liveness signal, so the panel cannot tell a plan still waiting from one already approved. Reminding anyway would reinstate exactly the bug this fixes. Decoupling "answerable from the panel" from "worth reminding about" would need a signal we don't currently have.

## Follow-up worth filing

`notify.sh` still has no test harness — `bash -n` and shellcheck only — so `is_question_event` and the trap fix are covered by nothing but the manual checks above. That's the gap flagged when #169 went in, and all three bugs in this PR lived exactly there. A ~30-line harness (fixture on stdin, a socket stub, assert `has_action` and `hook_pid`) would pin the lot.

Separately: `make reload` doesn't restart the app on my machine. `launchctl kickstart` fails with "Could not find service com.stackonehq.stack-nudge" and `|| true` swallows it, so it prints "reloaded" while the old binary keeps running.


notify.sh
```diff
@@ -350,20 +350,42 @@ agent_supports_decision() {
...
   [[ "$AGENT" != "claude-code" ]] && return 1
-  command -v jq &>/dev/null || return 1
   [[ -z "$HOOK_JSON" ]] && return 1
   local tool_name
-  tool_name=$(printf '%s' "$HOOK_JSON" | jq -r '.tool_name // empty' 2>/dev/null)
+  tool_name=$(hook_tool_name)
   case "$tool_name" in
-    AskUserQuestion) return 0 ;;
-    *)               return 1 ;;
+    AskUserQuestion|ExitPlanMode) return 0 ;;
```
StuBehan, line 387: `permission_context` has an AskUserQuestion case so the body isn't just the tool name, but ExitPlanMode falls through to the catch-all. That banner is now the only notification a plan gets - could we pull `.tool_input.plan` and truncate it the same way?

StuBehan (commented):
Fix reads right - I ran hook_tool_name/is_question_event with and without jq on the path and both classify ExitPlanMode as a question, so that fallback is load bearing and not just belt and braces. Trap bug is a good spot too.

One thing not in the diff: the README still promises reminders and a live menu-bar count for prompts waiting on you, and there's now a class of blocking prompt that gets neither. Worth a line there.

Two questions inline.

notify.sh
```diff
@@ -350,20 +350,42 @@ agent_supports_decision() {
...
+    pass' 2>/dev/null
+}
+
+# Agent-initiated question (multi-select / open prompt) or plan approval, not a
+# tool permission. Approving in the panel would write "allow" to the FIFO,
 # which CC interprets as "press Enter on the highlighted option" — making
 # it pick a default. By emitting has_action=false, panel Enter falls
 # through to focusing the editor so the user answers in the terminal.
+#
+# ExitPlanMode is the same shape: "allow" approves the plan outright, so the
```
StuBehan, line 376: Is there really no liveness signal though? We only register Stop and PermissionRequest - wouldn't a PostToolUse matcher on ExitPlanMode fire when the plan resolves either way? Not for this PR, but "isn't avoidable at this layer" is a strong thing to write down and then never revisit 🤔

PR #179, opened 2026-09-15
fix(panel): stop non-Events tabs inheriting the Events key bindings

hiskudin:
Found while planning the extensions work. Structural rather than reproducible on main today, but it's the kind that gets inherited silently.

## The problem

`panelHandlesKey` is a chain of per-mode blocks:

```swift
if nav.mode == .bootstrap { ... }
if nav.mode == .sessions  { ... }
if nav.mode == .outcomes  { ... }

// Events mode: ...
```

There is no guard before that last section. Anything that reaches it gets the Events bindings applied — and those include:

```swift
case KeyCode.returnKey: actOnSelected(approve: true)
```

So a mode with no block, or one that declines a key it doesn't want, could **approve a pending permission prompt from a screen that never showed it**.

## Why it doesn't bite on main

Every mode currently has a block, so nothing falls through today. The defect is the structure: a new tab inherits it by default, and you only find out if someone presses Return at the wrong moment.

It is already live on `feat/token-derby-tab` — the Derby block handles `R` and returns false for everything else, so `↑`/`↓`/`Return` leak into the Events store from the Derby tab.

## The fix

One guard, and the decision lifted to a named static so widening the set is a visible edit rather than an inline condition:

```swift
static func eventsOwnsKeyboard(_ mode: PanelMode) -> Bool { mode == .events }
```

Returning `false` rather than swallowing, so genuine app-level shortcuts still reach the responder chain from every tab.

## Testing

754 green. New `PanelKeyRoutingTests` asserts `.events` owns those bindings and that every other mode does not — listed explicitly rather than derived, because `PanelMode` gains an associated value when extension tabs land and `CaseIterable` won't be available then.

Manual: with a permission prompt pending, visit each tab and press Return. The prompt survives.


hiskudin:
Folded into #180 — same subject (tab/key plumbing), ~40 lines together, and the groundwork it was stacked under has no user-visible change to wait behind.

PR #181, opened 2026-09-15
fix(panel): group Settings into categories instead of one 51-row list

hiskudin:
Settings was one scrolling list of **51 rows in ten sections**, and the sections named control *types* rather than subjects:

- **"Toggles"** held ten rows spanning notification behaviour, panel behaviour and session naming
- **"Hotkey"** was a category of one
- **"Accent color"** is app-wide but sat under Widget

So the long list was half the problem and the grouping was the other half.

## Layout

A sidebar of eight subject categories with that category's rows beside it — the Usage tab's split, so the two-level keyboard model is one already in the app rather than a new one.

```
┌─ Settings ──────────────────────────────────┐
│ Notifications │ Paste Slack setup   token › │
│ Voice         │ Slack user     U0A8PS4B79D › │
│ Appearance    │ Send test message          › │
│ Usage         │ Slack notifications        ✓ │
│ Integrations ◀│ Notify when idle         30m │
│ Panel         │ Include message text       ✓ │
│ Events        │ GitHub PR links            ✓ │
│ Actions       │ Hide shipped               ○ │
└─────────────────────────────────────────────┘
```

Largest category is ten rows, against a list that was 51.

## Keyboard

- **↑↓** walk categories; **→** or **Enter** steps into the rows
- **Esc** steps back out to the categories, then hides the panel
- **⌘↑↓** jumps categories from inside the detail — Stu's suggestion. It replaces jump-to-first/last-row, which is a short trip now that a list is one category's worth.

←/→ keep cycling values, which is why Esc rather than ← is the way out — the Usage tab can use ← because nothing there consumes it.

## Two things worth a reviewer's attention

**Attention rows repeat.** Unwired agents, permissions and update stay pinned above the split and appear at the head of *every* category's row list, so they're reachable whichever category is selected. That's deliberate, not duplication.

**`index(of:)` has a new trap.** `settingsRows` is now the attention rows plus the selected category's, so `index(of:)` answers `0` for a row in a different category — silently selecting the wrong row rather than failing. Everything in the view is already scoped to the rendered category, so nothing is broken today, but `category(containing:)` now exists for anything reaching a row by identity. The updated tests use it.

## Testing

781 tests, 2000 assertions.

`SettingsRow` gains `CaseIterable` for one reason: a completeness test that proves the hand-regrouping of 51 rows dropped nothing. Every case must be an attention row, a voice row behind the cached-model branch, or in exactly one category. I verified it by deleting `.quit` from Actions and watching it fail — without that test, a dropped row would simply vanish from Settings with nothing to catch it.

Four existing tests encoded the flat list; one force-unwrapped `pinPanel`'s index and crashed the runner with a SIGTRAP once it moved categories. They now assert against `rows(in:)` for the relevant category, and the index⇄row round-trip walks every category rather than whichever happens to be selected.

Manual: every category opens, ⌘↑↓ walks them from inside the detail, Esc steps out then hides, and value cycling still works on ←/→.

## Note on the version bump

Raised as `fix:` per request. That produces a **patch** bump (1.34.0 → 1.34.1); `feat:` is what would give a minor. Flagging in case patch isn't what was intended — the change is user-visible and arguably a feature.


panel/PanelNav.swift
```diff
@@ -1035,29 +1045,82 @@ final class PanelNav: ObservableObject {
...
+    }
+
+    // The first row the detail actually renders. Index 0 is an attention row
+    // whenever one exists, and those render above the split — so selecting 0
+    // put the selection on a row the pane wasn't showing, and Enter fired it.
+    // With an update pending that starts the updater; with unwired agents it
+    // rewrites every detected agent's hook config. ↑ from here still walks up
+    // into the banners, which is where they're drawn.
+    func selectFirstCategoryRow() {
+        selectedSettingIndex = settingsAttentionRows.count
```
StuBehan, line 1119: This parks on `settingsAttentionRows.count`, but that count moves at runtime - `UpdateChecker` sets `nav.updateAvailable` off a background check and `missingPermissions` gets set async too. So if either fires while Settings is open and sitting here, the selection slides onto the last attention row, which is drawn up in the banner, and ⏎ there starts the updater or rewrites every detected agent's hooks. Same scenario as `testEnteringACategoryDoesNotSelectAnInvisibleAttentionRow`, just reached by a timer instead of a keypress. Re-clamp when the attention rows change, or hold the selection by row identity rather than index? (same shape on main so not strictly yours, but this feels like the place to close it)

panel/Settings.swift
```diff
@@ -597,25 +467,206 @@ struct SettingsView: View {
...
+            .frame(maxWidth: .infinity, alignment: .leading)
+            .padding(.horizontal, 8)
+            .padding(.vertical, 5)
+            .background(
+                RoundedRectangle(cornerRadius: 6, style: .continuous)
+                    .fill(selected ? Color.accentColor.opacity(0.12) : .clear)
+            )
+            .contentShape(Rectangle())
+            .onTapGesture {
+                nav.settingsCategory = category
```
StuBehan: Is jumping straight into the detail deliberate here? I clicked the sidebar so I'd expect to still be on the sidebar, and now ↑↓ move rows instead of categories.

panel/Settings.swift
```diff
@@ -597,25 +467,206 @@ struct SettingsView: View {
...
+                        // They index ahead of the category's rows, so this is
+                        // also the order the keyboard walks.
+                        settingsBanners
+                        detailRows
+                    }
+                    .padding(.horizontal, 14)
+                    .padding(.vertical, 12)
+                    .background(ThinScrollers())
+                }
+                .onChange(of: nav.selectedSettingIndex) { newIndex in
```
StuBehan, line 616: Switching category leaves `selectedSettingIndex` at the same value every time, so this never fires and nothing tells the ScrollView to go back to the top - does the offset persist? Scroll into Integrations, Esc, ↑ to Notifications. Haven't run it so I might be wrong about what SwiftUI does there 🤔

StuBehan (commented):
@hiskudin this is a big improvement, and the completeness test is the bit I like most - `CaseIterable` plus `testEverySettingsRowHasAHome` means a dropped row fails rather than quietly disappearing, which was unfindable before. Few comments inline.

On the version bump - it's user-visible so I'd go `feat:` and take the minor. Fine to leave it as a fix if you'd rather it rides with 1.34.1.

panel/PanelNav.swift
hiskudin: Confirmed and fixed. Both triggers are real: `updateAvailable` is set from a repeating `Timer` in `UpdateChecker.swift:86` and `missingPermissions` from an async callback at `PanelNav.swift:1285`, so either lands while Settings is open with nobody touching the keyboard.

I went with anchoring rather than re-clamping — re-clamping papers over it, holding the row identity actually fixes it. The selection now remembers *which row* it is on and re-derives its index whenever the attention rows change, so it follows its row instead of its number.

The case worth calling out is a row that disappears entirely (the update installed, the agent wired): the selection falls back to the first row of the category rather than to whatever inherited its index, because inheriting an index is exactly how a keypress meant for a toggle ends up starting an updater.

Six regression tests, all mutation-verified. One of them was too weak on the first pass — my vanished-row test passed against a mutant that kept the old index, because in that scenario both paths happened to land on the same row. Rewritten to three attention rows with the selection on the first, where keeping the number puts it on the update banner.

You are right that it is the same shape on main; this branch is what makes it systematically reachable, since parking exactly at the boundary is now the normal state.

panel/Settings.swift
hiskudin: Not deliberate, and you are right that it is inconsistent. The Usage tab is the precedent — clicking a row in the list picks it, and →/Enter is what steps inside — so focusing the detail on a sidebar click silently repurposed ↑/↓ from categories to rows on a click you made *to change category*. Dropped the line.

panel/Settings.swift
hiskudin: You were right on the mechanism, and it does persist. `selectFirstCategoryRow()` sets the same number every time, so `onChange(of: selectedSettingIndex)` never fires, and the ScrollView keeps its offset because its own identity is stable across the content swap — a short category opened part-scrolled.

Added an `onChange(of: settingsCategory)` that resets to the top.

One thing I checked while I was in there, since it would have made your suspicion much worse: `scrollTo(newIndex)` does match real ids — the rows are tagged `.id(nav.index(of: id))` — so scroll-into-view itself was working.

PR #182, opened 2026-09-15
fix(panel): stop the claude usage probe timing out

StuBehan:
The usage tab keeps flashing "Couldn't refresh - run `claude /usage` to check your session", but nothing's actually broken - the probe is just timing out.

`claude --print /usage` is a client-side intercept, needs no MCP, but every spawn still boots the whole MCP config (60+ servers on my setup) before it runs. That startup tail is variable and under load it pushes the whole call past our 10s timeout, which we read as a hard-fail and surface as an error - even though the CLI would have answered fine a second or two later. Intermittent because it straddles the 10s line: quiet machine squeaks under, busy machine (4+ parallel sessions) tips over.

Two changes:
- `--strict-mcp-config` on the /usage call - loads zero MCP servers, which is right since the intercept needs none, and stops 61 servers spinning up ~1.4k times/day for nothing,
- timeout 10 -> 20s - strict-mcp-config alone wasn't enough, I still saw 10.96s on a busy machine with MCP off. The intercept's own local-session scan grows under load, so 20s gives it headroom rather than sitting right on the edge.

Timing: ~6.5-13s before, ~6.5s with the MCP tail gone (though still up to ~11s when the box is busy, hence the 20s). The envelope's own `duration_ms` is ~5s, the rest was startup.

No new test - it's two literals into a shell-out and the parse path (which is covered) is unchanged, so there's no seam to assert without a refactor. Confirmed by hand that `--print --strict-mcp-config --output-format json /usage` still returns a parseable envelope, and `make test-without-xcode` is green (771 tests).


hiskudin:
Reviewed, and I reproduced the parts I could. The diagnosis holds and I'd take this — two things to flag, one of them a compatibility risk in this change, and one unrelated bug I tripped over while testing it.

## Verified

**`--strict-mcp-config` produces byte-identical tier output.** Ran both forms against my account:

```
with    --strict-mcp-config : duration_ms=954, 3 tier lines, wall 3.4s
without --strict-mcp-config : duration_ms=936, 3 tier lines, wall 3.0s
same tier lines: True
```

So the claim that the intercept needs no MCP checks out — the parsed result is unchanged.

**I can't reproduce the speedup**, because I have almost no MCP servers configured: 3.4s vs 3.0s, i.e. noise. That's consistent with your explanation rather than against it — the win is proportional to config size, and your 60+ server setup is where it shows. Worth noting for anyone else who tries to confirm it and sees nothing.

## The risk I'd want covered before merging

**An older `claude` without the flag turns this into a permanent failure.** I checked what an unsupported flag does:

```
$ claude --print --definitely-not-a-real-flag --output-format json /usage
exit=1
error: unknown option '--definitely-not-a-real-flag'
```

Non-zero exit, no JSON on stdout. That routes through `parseEnvelope` → `.hardFail` → *"Couldn't refresh — run `claude /usage`"*, forever, on every poll. So for anyone on a CLI predating `--strict-mcp-config`, this fix causes exactly the symptom it's fixing — and worse, because the current bug is intermittent while that one is permanent.

Most people auto-update, so the blast radius is probably small. But there's no minimum-version check anywhere in the probe, and the failure is silent and indistinguishable from a real outage.

Cheapest fix that keeps the win: on hard-fail, retry once without the flag and remember the result for the process lifetime. About the same size as the change itself, and it turns an unsupported flag into one slow first poll rather than a dead tab.

## We fixed two adjacent causes in #180

Same symptom, different roots — these stack rather than conflict:

- The freshness label read the **global** `quotaLastUpdated`, stamped by any client. A failing Claude showed "Updated 12s ago" because Codex had just polled — which is why the error looked spurious even when it was accurate. Now per-client.
- That message also fires on the **`cliMissing`** path, where the binary simply wasn't found: `ProcessOutput.claude()` walks four absolute paths, so volta/bun/mise/asdf installs miss. Those users were told to run a command that works fine in their terminal. Now names the real fault, plus shim paths and `STACKNUDGE_CLAUDE_PATH`.

Your timeout case is the third, and on a machine with real MCP config probably the most common of the three.

## Unrelated bug I hit while testing this

My `/usage` output now includes a tier this repo drops on the floor:

```
Current session: 13% used · resets Sep 15 at 7pm
Current week (all models): 25% used · resets Sep 19 at 3am
Current week (Fable): 0% used · resets Sep 19 at 3am     <- dropped
```

`parseResultText` switches on four exact names (`session`, `week (all models)`, `week (Opus only)`, `week (Sonnet only)`) with `default: continue`, so **`week (Fable)` is silently discarded** and never reaches the Usage tab. Any per-model tier Anthropic adds disappears the same way, with no error and nothing failing.

Not yours to fix here — raising it because this file is the one you're in, and because it's the same shape as the `#{host}` format-string gap from #169: two halves of a contract where only one half is pinned.

hiskudin:
Correcting myself on the last point — I said the `week (Fable)` drop wasn't yours to fix here, which wasn't my call. You're in that file; take it in this PR if you'd rather have it done in one go.

The only argument for keeping it out is that this PR is a clean +10/-2 and the fix changes shape rather than literals — `parseResultText` would need to stop mapping four exact names onto four named slots and carry the weekly tiers it actually finds. `windowLength(forTier:)` already does `hasPrefix("week")`, so that half is fine.

Either way works. Say if you'd rather I picked it up separately.

StuBehan:
@hiskudin thanks for actually running both forms, that's the bit I couldn't prove on my own machine.

Taking the fallback. You're right that it swaps an intermittent failure for a permanent one and that's the worse trade, even if most people auto-update. Doing it the way you described - hard-fail retries once without the flag, remember the answer for the process lifetime - so a stale CLI costs one slow poll instead of a dead tab.

On the Fable tier, going to do that separately. It's a bit worse than you found actually: `foundAny` is set before the switch, so a `/usage` that only has unrecognised tiers comes back `.ok` with an all-nil snapshot rather than `.softFail`. Tab gets a successful, empty Claude row and no error at all. If they ever rename `week (all models)` that's the failure mode. `windowLength(forTier:)` already does `hasPrefix("week")` so the weekly half is easy, but it's a shape change to `parseResultText` and I'd rather not hang it off a +10/-2.

Will raise it - happy for you to pick it up if you want it.

StuBehan:
Good shout, the older-cli case is the real risk. Pushed a guard: runUsage sends --strict-mcp-config, and if the cli rejects it (your exit-1-empty-stdout) it retries once without the flag and remembers that for the process. Only latches off if dropping the flag actually gives output, so a real outage doesn't kill the speedup, and it leaves timeouts alone (retrying there just reloads all the mcp servers, ie the slow path we're fixing). Tests cover it.

On week (Fable), yeah its real, my own account drops it too. Gonna take it separately so this stays the small fix - it's the same default: continue you binned from tabLabel in #180. Happy to pick it up or leave it yours, lmk?

PR #183, opened 2026-09-15
feat(panel): run and render out-of-process extensions

hiskudin:
PR 3 of the [extensions plan](https://github.com/StackOneHQ/stack-nudge/pull/180). Branches off `main`, independent of #181.

## What this is

An extension is a directory under `~/.stack-nudge/extensions/<id>/` holding a `manifest.json` and an executable. The panel spawns it, reads one JSON view document off stdout, and renders it into a tab. No native code, no resident process — a spawn per invocation, exactly like `ClaudeCliQuotaProbe`.

Nothing downloads yet. Extensions are read from disk only; distribution and the Settings browser are PR 4.

## The schema

This is the part worth the review attention, because it becomes a contract the moment anything ships against it.

```json
{"schema":1,"state":"ok",
 "header":{"title":"League","badge":{"text":"LIVE","tone":"success"},"trailing":"5h44m"},
 "rows":[{"id":"h1","lead":"1","title":"black & white","subtitle":"Yashika","value":"14.5M",
          "footnote":"36K/15m","track":{"fill":0.52,"ghost":0.60,"tint":"#7FD1B9"},
          "ornament":{"kind":"sprite","fps":7,"anchor":"fill-edge",
                      "palette":{"H":"#FFFFFF"},"frames":[["..HH"],["HH.."]]},
          "actions":[{"id":"open","label":"Open","key":"return"}]}],
 "actions":[{"id":"refresh","label":"Sync now","key":"r"}]}
```

The primitives are deliberately generic, because the host renders them without knowing what any of them mean. **"A grid of coloured cells" is a primitive; "a horse" is not** — which is what lets the Derby's sprite survive the process boundary in PR 5. `track.ghost` is the pace bar's shape from #176, generalised.

`schema` is refused rather than coerced when it isn't 1, and additive fields are the only compatible change.

## Failures split three ways

They mean different things, so they read differently:

| | |
|---|---|
| **missing** | script gone or not executable — a broken installation |
| **transient** | timeout or non-zero exit — hold the last good document, mark it stale |
| **malformed** | output didn't parse — a bug in the extension, surfaced as one |

A non-zero exit is transient *even when stdout parsed*, because a script that failed halfway may have printed a partial document.

## What an extension can't do

This is curation-backed, not sandboxed — the script runs as the user. But within that:

- **Key bindings are requested, not granted.** The host maps letters, digits and Return. Esc, every arrow and every ⌘ combination stay the host's. An action requesting one keeps its button and loses its shortcut.
- **Unbound keys in an extension tab are swallowed**, never passed on — so a document can't reach the Events handler and answer a permission prompt. (The general fall-through fix landed in #180; this is the tab holding up its end.)
- **The child's environment replaces ours**, so an extension sees only the `STACKNUDGE_*` keys its manifest declared. Declaring `PATH` or `AWS_SECRET_ACCESS_KEY` gets it nothing.
- **Ids and run paths are validated before they reach a path.** `../../evil` is refused as an id, as a run path, and again at spawn time — with a canary planted outside the directory to prove the last one stands on its own.
- **One invocation in flight per extension**, so two presses can't race to publish.

## Deviations from the plan

- **No `ProcessOutput.readJSON`.** What was actually needed was the exit status, so `read` is now a thin wrapper over a new `run` returning `(status, output)`. Behaviour-preserving for existing callers; JSON parsing stays in the runtime where the three-way failure split lives.
- **`requires` is parsed but not yet enforced.** The plan puts the check at install time, which is PR 4. Writing it here would have meant shipping a checker with no installer to call it.

## A test-harness fix, found by mutation testing

Two guards survived mutation: a manifest claiming an id other than its own directory, and an escaping run path. Both tests were passing against an **empty world** — the no-Xcode runner calls `setUp()` but never `setUpWithError()`, which the shim has always declared, so every filesystem fixture silently never ran. A throwing test body was also invoked with `try?`, reporting a failed fixture as "ok".

Both are fixed in `scripts/run-tests-without-xcode.sh`. No other test in the tree used `setUpWithError`, which is why it went unnoticed. After the fix, all four guards above are caught by mutation.

## Verification

`853 test(s), 2130 assertion(s), 0 failure(s)` locally; `./build.sh` and `make typecheck-tests` clean.

Spawn tests run real scripts in a real temp directory — a stubbed `FileManager` would have tested the stub's idea of executability rather than the kernel's. Covered: valid document, malformed JSON, non-zero exit with parseable stdout, missing interpreter, missing and non-executable run files, empty stdout, argv round-trip, environment isolation, traversal canary.

Mutation-verified: id-matches-directory, the spawn traversal guard, the non-zero-exit split, the config namespace filter, the key allowlist, focus gating, selection reconciliation, and the in-flight gate.

Manually: a demo extension exercising every primitive renders, animates, and round-trips a row action.

---

# Adversarial review round

Five reviewers (security, schema contract, concurrency, rendering, tests) found ~25 defects, most reproduced rather than argued. Fixed across two commits — the `ProcessOutput` work is separate so it can be read alone.

## The one that mattered most

`readDataToEndOfFile` on a pool thread **cannot be abandoned**. EOF needs every holder of the write end to close it, and a child that backgrounds anything hands that end to a grandchild the host never sees. So a script that printed a valid document and exited in milliseconds was reported as a **12-second timeout**, leaked a thread and its pipe each time, and at ~64 leaks **starved the shared global utility pool** — the same pool as the session poll and the quota probes.

That is the wedge class this app has been bitten by three times, and the reason extensions spawn per invocation. It was reintroduced one level down, on the one input that is third-party.

Now a cancellable read source, with a one-second grace on the post-exit drain. Measured: clean path unchanged at 0.01s, leaky case `nil@12s` → `status 0 @1.01s`, descriptors flat across runs. Signature and the `nil` contract unchanged, so all 14 callers are behaviour-compatible.

Alongside: timeouts now signal the process **group** (read back via `getpgid`, never our own) instead of leaving orphans; `terminationStatus` after a signal is the signal number, so `Completion` carries `signalled` rather than rendering SIGSEGV as "exited 11"; and the `Pipe`'s descriptors are closed explicitly, because its `FileHandle`s measurably do not deallocate promptly — one FIFO leaked per call.

## Containment

The run-path guard compared a **lexically** standardized path, which proves only that the *spelling* starts inside the package — `isExecutableFile` and `Process` both follow symlinks. A `vendor -> /usr/bin` entry with `"run": "vendor/whoami"` passed the guard and spawned `/usr/bin/whoami`.

This matters *because* the model is curation rather than sandboxing: the guard is what a reviewer relies on when they read a manifest and conclude the executed bytes are the ones in the package, and a tarball carries symlinks fine. Both sides are resolved now — including the root, since a symlinked home is ordinary and resolving one side would break every legitimate install.

Also: the environment comment understated the ceiling. macOS attributes AppleEvents/FDA to the **responsible process**, which a child inherits and Foundation gives no way to disclaim — so an extension can do what the user can do *without being asked*. Corrected, because curation is the only control and that comment is what a reviewer decides against.

## Contract

- **Key bindings counted graphemes.** `"a"` + U+FE0F, a ZWJ, a combining mark and `1️⃣` were all granted — a binding the key handler can never match, plus a footer hint advertising it. Counted in scalars now; the test builds them with explicit `\u{}` escapes rather than trusting this file's normalisation.
- **Refusals were discarded.** A newer-schema manifest produced no tab, no message, no log — indistinguishable from "wrong folder", and it made the versioning policy unfalsifiable. Reported now.
- **Row leniency is per row**, not per field: one item missing a `title` used to blank the whole pane.
- **Sprite rows, columns and frame count capped** — capping fps alone was pointless when a million-column frame parses. Tab labels bounded too.
- **Exports `STACKNUDGE_SCHEMA`** — free now, impossible to retrofit.

## Rendering

Two extension tabs **shared one view identity** (a ViewBuilder `switch` case is one branch; the associated value isn't identity), so switching between them never fired `onAppear` and the second tab stayed permanently unfetched. A `.broken` fetch over a live document showed stale numbers **with no marker at all** — the exact failure the three-way split exists to prevent — and the marker and spinner only existed inside the *optional* header. Plus: keyboard selection now scrolls into view; sprites are sized and anchored from one shared maximum and clipped to their band; the track bar is centred like the `paceBar` it claims to copy; an ornament on a track-less row renders; a palette colour that doesn't parse falls back.

Extension colour is **clamped toward readable** against the current appearance (blended, so hue survives) — a well-formed `#FFFFFF` was an invisible bar in light mode. Accessibility follows the Usage tab: the track carries a label and value, the selected row carries the trait, the decorative sprite is hidden.

## Deliberately deferred

Contract *design* questions, not bugs — worth deciding rather than defaulting:

- `state:"error"`/`"empty"` **with rows present** discards them. "Degraded, but here's what I got" is a real want, but it changes what `error` means.
- Four enum-ish fields, four unknown-value policies, all case-sensitive while `key` is lowercased.
- Message quality: `"schema":1.5` reports "not valid JSON"; `schema:0` reads "needs manifest schema 0"; decoder paths leak `rows.Index 0.track.fill` into the pane.
- Duplicate JSON keys are **first**-wins in Foundation, opposite to Python/JS. Documentation, not code.
- `finish` can resurrect a pane `load()` dropped — latent, since `load()` is launch-only until PR 4.
- The `@MainActor` removal leaves the main-thread invariant unenforced: no strict-concurrency setting anywhere, so `swift test` catches nothing. No off-main mutation exists today; every caller was traced.
- Whether sprite timelines keep ticking while the panel is ordered out — the one thing the reviewer refused to assert without Instruments.

## Verification

`905 test(s), 2233 assertion(s), 0 failure(s)`; all six CI checks green.

Every new guard **mutation-verified**, in a dedicated `git worktree` — after running mutations in the shared tree broke four concurrent reviewers' builds and nearly produced a phantom finding.

Seven mutants survived a first pass and are now covered. Two from my own round: the extensions root behind a symlink, and the row-title guard. Five the test reviewer found, each one green with the guard deleted — `isDue`'s non-positive interval check, `handle(key:)`'s in-flight guard (the existing test covered the *document* half of the same guard and never the busy half), `installed()`'s sort (two fixtures weren't enough to force it — the filesystem returned them already ordered), and the badge-tone and sprite-anchor fallbacks, where every fixture sent a valid value so the defaults could have been anything.

The renderer's pure statics had **no tests at all** — colour parsing, the contrast clamp, sprite geometry. Writing them found a real bug: the clamp read `NSApp`, an implicitly unwrapped optional that is nil outside a running application, so it was a crash under the test runner and briefly at launch rather than a wrong colour. It also corrected the clamp itself — `NSColor.blended` mixes through its own colour space and overshot, landing white at 0.795 luminance against a 0.75 ceiling. Solved in luminance terms now, which is exact and still moves every channel together, so hue survives.

---

# Stu's review round

All seven comments were real. Fixed in `1ad8fab`.

## The drain grace was losing data, and the fix is better than the flag

Stu caught that the rewrite turned a *failure* into a *wrong answer*: the old code returned `nil` when the drain outran its budget, mine handed back whatever had arrived as `status 0`. A clipped `ps -axo args=` parses perfectly well as a **shorter session list**, so under the very pool starvation this work exists to fix, sessions would have vanished silently.

His question — can we tell "exited and saw EOF" from "exited, grace expired"? — pointed at something stronger than the bit I was discarding. **Once the child has exited, everything it wrote is already in the pipe buffer**, so a read that reaches `EAGAIN` has seen all of it whether or not EOF arrives. Draining to quiescence after exit makes completeness answerable where EOF alone could not distinguish "complete, helper idle" from "incomplete, helper still writing".

So for `SessionStore` the truncation is **structurally gone**, not merely reported. `Completion.truncated` remains for what is still genuinely unknown — the drain not quiescing inside its grace, and output past the new 8MB ceiling — and `read` returns `nil` for both, restoring what `nil` has always meant to its 14 callers.

## A prefix is not a secret filter

`STACKNUDGE_SLACK_BOT_TOKEN` lives in the config namespace, and `SlackCredentials` deliberately leaves it in plaintext when the Keychain is locked (scrubbing a token the Keychain refused would destroy it) — exactly the window in which a manifest naming it would walk off with a live bot token.

Extensions now have their own **`STACKNUDGE_EXT_`** namespace, so nothing the app stores is nameable. A deny-list would need extending every time a config key is added, which is the same bug deferred. Enforced at manifest parse *and* in the environment builder; there is a regression test naming `SlackCredentials.configTokenKey` directly.

## The rest

- **`signalled` was a dead flag** — set, tested twice, read nowhere, so a segfaulting extension still said "exited 11". `classify` reads it now, and the test asserts the *rendered message*, since asserting the flag is what let it be dead.
- **Rows, actions and text fields capped**; stdout bounded at 8MB. Bytes past the ceiling are read and discarded rather than left in the pipe — abandoning it would block the child and turn its overrun into a timeout.
- **`dup` failure** no longer orphans the child, which matters because fd exhaustion is this function's own failure mode.
- **The `isRunning` gate** is restored. Stated plainly: this one has **no test** — mutating it away leaves the suite green because the window is microseconds wide and only on a timeout. It is in on consistency, not evidence.

## On the harness fix

Stu's count is exact: **23 throwing tests across 7 files** on `main` (10 `BootstrapTests`, 6 `HotkeyTests`, 2 each `SlackCredentialsTests` and `ClaudePathOverrideTests`, 1 each in three more) were running under `try?` and could not fail. Nothing was hiding, but that is the gate everyone runs locally.

Worth saying how it surfaced: two of my own guards **survived mutation testing** because their fixtures were built in `setUpWithError` and silently never ran — the tests were passing against an empty world.

`917 test(s), 2263 assertion(s), 0 failure(s)`; all six checks green. Every new guard mutation-verified in a worktree except the `isRunning` gate, noted above.


StuBehan (commented):
@hiskudin this is solid, and the `ProcessOutput` rewrite is the part I care about most because it isn't really about extensions - `SessionStore`, the terminal integrations and both quota probes all go through it. Cancellable read source instead of `readDataToEndOfFile` on a pool thread is the right fix, and the grandchild-holds-the-write-end reasoning is correct.

Comments inline. The drain grace one is the one I'd most want another look at.

panel/ProcessOutput.swift
```diff
@@ -56,30 +88,110 @@ enum ProcessOutput {
...
+
+        // One deadline for the whole call. The old shape waited `timeout` for
+        // exit and then a further `timeout` for the drain, so the worst case was
+        // double what the budget — and the user-facing message — claimed.
+        let timedOut = exited.wait(timeout: .now() + timeout) == .timedOut
+        if timedOut {
+            terminateTree(task, exited: exited)
+        }
+        // Either way, give whatever is still open a bounded grace, then stop.
+        _ = drained.wait(timeout: .now() + drainGrace)
```
StuBehan: This is the one I'd most want another look at, and it's the non-extension callers I'm thinking about. Old path returned nil if the drain outran the budget, this hands back whatever it got as status 0 after a second. Right call for the grandchild case. But under the pool starvation this whole change exists to fix, a slow drain gives `SessionStore` a truncated `ps -axo args=` that parses perfectly well as a shorter session list - silently dropping sessions is worse than failing outright. Can we tell "exited and saw EOF" apart from "exited, grace expired"?

panel/ProcessOutput.swift
```diff
@@ -56,30 +88,110 @@ enum ProcessOutput {
...
+        // forever, leaking a thread and the pipe with it. At around 64 of those
+        // the global utility pool is starved and unrelated work stops being
+        // scheduled: the sessions poll, the quota probes, everything.
+        //
+        // The fd is duplicated so the source owns a descriptor it can close in
+        // its cancel handler; the Pipe's own FileHandle closes the original. A
+        // dup shares the file description, so reading from it drains the same
+        // pipe.
+        let readFD = dup(outPipe.fileHandleForReading.fileDescriptor)
+        guard readFD >= 0 else { return nil }
```
StuBehan: If `dup` fails here we return nil with the child still running and `exited` still entered, so an orphan plus a leaked group. Only bites on fd exhaustion, which is the case this is about 🤷

panel/ProcessOutput.swift
```diff
@@ -56,30 +88,110 @@ enum ProcessOutput {
...
+    // just the child. Foundation spawns into a new group, so a script that
+    // backgrounds `curl &` and then hangs used to leave the curl running after
+    // every timeout, forever. The group is read back rather than assumed, and
+    // our own is never signalled: killpg on the wrong id would take out the app.
+    //
+    // Waiting on the termination group rather than polling isRunning, so a child
+    // that refuses to die costs a bounded wait instead of a spinning thread.
+    private static func terminateTree(_ task: Process, exited: DispatchGroup) {
+        let pid = task.processIdentifier
+        let group = getpgid(pid)
```
StuBehan, line 269: We lost a guard the old code had - it was `task.terminate()` then `kill(SIGKILL)` gated on `task.isRunning`. Now we signal unconditionally, and once the child has exited and been reaped `getpgid` gives -1 so we fall through to `kill(pid, ...)` on a pid that could have been recycled. Microseconds wide and only on a timeout, I know, but the code already goes to the trouble of never signalling our own group so it feels odd to leave this one.

panel/ExtensionRuntime.swift
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    // A non-zero exit is transient even when stdout parsed, because a script
+    // that failed halfway may well have printed a partial document — believing
+    // it would show stale rows as current. Empty stdout on a clean exit is the
+    // same call: the extension ran, said nothing, and the last good document is
+    // a better thing to show than a blank pane.
+    static func classify(_ completion: ProcessOutput.Completion) -> Fetch {
+        guard completion.status == 0 else {
+            return .transient("exited \(completion.status)")
```
StuBehan: `signalled` never gets read - set in `run`, asserted in two tests, and that's it. So a segfaulting extension still reports "exited 11" here, which is the thing the flag exists to stop. One line?

panel/ExtensionRuntime.swift
```diff
@@ -0,0 +1,195 @@
...
+            "PATH": "/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin",
+            "HOME": home,
+            "STACKNUDGE_EXTENSION_ID": manifest.id,
+            // The one thing an extension needs to adapt to an older host. Free
+            // now and impossible to retrofit: without it "additive only" has no
+            // migration path, because a script has no way to ask what this host
+            // can read before it prints.
+            "STACKNUDGE_SCHEMA": "\(ExtensionManifest.supportedSchema)",
+        ]
+        for key in manifest.config where key.hasPrefix("STACKNUDGE_") {
```
StuBehan: `STACKNUDGE_SLACK_BOT_TOKEN` is in this namespace (`SlackCredentials.swift:25`). Normally it's in the Keychain, but the migration deliberately leaves the plaintext in config when `store(botToken:)` fails on a locked keychain, and that's exactly the window where a manifest declaring that key walks off with a live bot token. `STACKNUDGE_SLACK_MEMBER_ID` same. Curation is the model and I'm fine with that, but a prefix isn't a secret filter and it widens silently every time we add a config key. Allowlist, or give extensions their own `STACKNUDGE_EXT_` namespace?

panel/ExtensionDocument.swift
```diff
@@ -0,0 +1,298 @@
...
+        }
+        // An unrecognised state is not a reason to throw the document away —
+        // rows are the substance and they parsed. Treat it as ok, which is the
+        // reading that shows the most.
+        let state = decoded.state.flatMap(State.init(rawValue:)) ?? .ok
+
+        // Rows sharing an id would collapse under ForEach and make actions
+        // ambiguous — the action payload names a row by id and nothing else.
+        var seen = Set<String>()
+        let rows = decoded.rows?.compactMap { row -> Row? in
```
StuBehan: Nothing caps the row array - sprite frames, rows and columns are all capped but this isn't, nor the strings, nor the actions list, nor stdout (`run` appends to an unbounded `Data`). So a script printing a few GB inside the 12s budget takes the panel with it. Same class as the sprite caps you closed on purpose. (`LazyVStack` saves the rendering but we've parsed a million `Row`s by then anyway)

scripts/run-tests-without-xcode.sh
```diff
@@ -65,9 +70,9 @@ for source in "$TESTS_DIR"/*.swift; do
     /^    func test[A-Za-z0-9_]*\(\)/ {
       match($0, /test[A-Za-z0-9_]*/)
       fn = substr($0, RSTART, RLENGTH)
-      call = (index($0, "throws") > 0) ? "try? " : ""
+      call = (index($0, "throws") > 0) ? "try " : ""
       if (cls != "") {
-        printf "        run(\"%s.%s\") { let c = %s(); c.setUp(); %sc.%s(); c.tearDown() }\n", \
+        printf "        run(\"%s.%s\") { let c = %s(); try c.setUpWithError(); c.setUp(); %sc.%s(); c.tearDown(); try c.tearDownWithError() }\n", \
```
StuBehan, line 75: This deserves more shouting about than it's getting. I checked main - nothing used `setUpWithError`, but there are 23 throwing tests across 7 files that have been running under `try?` and unable to fail. Green on your branch so nothing was hiding, but that's the gate we all run locally.

panel/ProcessOutput.swift
hiskudin: Right, and the fix turned out to be better than the flag you asked for — your question pointed at something stronger than the bit I was throwing away.

Once the child has exited, everything *it* wrote is already in the pipe buffer. So a read that reaches `EAGAIN` has seen all of it, whether or not EOF ever arrives. Draining to quiescence after exit makes the question answerable, where EOF alone genuinely could not tell "complete, helper idle" from "incomplete, helper still writing".

So for `SessionStore` the truncation is now structurally gone rather than reported: if `ps` is still running at the deadline we time out and return nil as before, and if it exited then its output is complete by the argument above. A background helper holding the write end is harmless rather than merely tolerated.

`truncated` stays for what is still genuinely unknown — the drain not going quiet inside its grace, and output past the new 8MB ceiling — and `read` returns nil for both, which restores what nil has always meant to its 14 callers.

Two tests pin the distinction: an idle helper must *not* mark the output partial, and a chatty one must not either (the child's own line is complete; the helper's bytes afterwards are a different process's).

panel/ProcessOutput.swift
hiskudin: Fixed — terminates the tree before returning. And you're right that it is not hypothetical: fd exhaustion is this function's own failure mode, so it is precisely when `dup` will fail.

panel/ProcessOutput.swift
hiskudin: Agreed, and the inconsistency is the part that convinced me. `signalAll` now returns early unless `task.isRunning`, and skips when `getpgid` gives -1 rather than falling through to `kill(pid, ...)`.

Being straight about the evidence: this one has no test. Mutating the guard away leaves the suite green, because the window is microseconds wide and only on a timeout, and I could not construct it deterministically. Everything else in this round is mutation-verified; this is in on the consistency argument, not on a test.

panel/ExtensionRuntime.swift
hiskudin: Caught me — that is exactly the dead-flag criticism I spent the previous review round levelling at other people's code. `classify` reads it now, and the test asserts the *rendered message* rather than the flag, since asserting the flag is what let it be dead in the first place.

panel/ExtensionRuntime.swift
hiskudin: You're right, and "a prefix isn't a secret filter, and it widens silently every time we add a config key" is the part that settled it — a deny-list would be the same bug deferred.

Extensions now have their own `STACKNUDGE_EXT_` namespace, so nothing the app stores can be named at all. Enforced at manifest parse as well as in the environment builder, so a stray key is a refusal you read in the PR rather than an empty variable a script discovers at runtime. `STACKNUDGE_EXTENSION_ID` and `STACKNUDGE_SCHEMA` are renamed to `STACKNUDGE_EXT_ID` / `STACKNUDGE_EXT_SCHEMA` — the old name sat one character from the new prefix and read like it should match it.

There is a regression test that names `SlackCredentials.configTokenKey` directly and asserts it cannot be reached by either path.

panel/ExtensionDocument.swift
hiskudin: Fixed — rows, actions and the text fields are all capped now, and stdout is bounded at 8MB in `ProcessOutput`.

One thing worth stating: the bytes past the ceiling are read and *discarded* rather than left in the pipe. Abandoning it would block the child on a full buffer and turn its own overrun into a timeout instead of a truncation.

Rows are capped after the malformed ones are dropped, so one bad row doesn't cost a good one its place in the list.

scripts/run-tests-without-xcode.sh
hiskudin: Fair, and I checked your number: 23 throwing tests across 7 files — 10 in `BootstrapTests`, 6 in `HotkeyTests`, 2 each in `SlackCredentialsTests` and `ClaudePathOverrideTests`, 1 each in three more. Exactly right.

I have moved it out of the commit message and into the PR body. It also deserves saying that this is how the gap was found at all: two of my own guards survived mutation testing because their fixtures were built in `setUpWithError` and silently never ran, so the tests were passing against an empty world.

PR #190, opened 2026-09-18
feat(extensions)!: the Token Derby, and a settings form for extension config

hiskudin:
Three things, in the order they depend on each other: the Token Derby as an
extension, a way to configure an extension, and a place to find both.

## The Derby

It was written as a built-in Swift tab and never merged. This is that tab
written against the extension runtime, which is what it should have been — it
talks to a third-party service and belongs to one organisation, and neither of
those has any business being compiled into the app. Nothing is deleted; none of
the old branch is on `main`.

Everything host-shaped went away in the move. The colour legibility clamp, the
sprite geometry and the polling schedule are the host's now, declared in
`manifest.json` rather than coded in the extension. What is left is the part
that is about racing.

Checking it against the live service changed three things:

- **The race-selection rule now matches the derby's own website** — newest
  live, else newest finished, else the *soonest* pending, which sorts the other
  way because a race that hasn't started is interesting for being next rather
  than recent. An extension that quietly picked a different race from the page
  it mirrors would read as stale data.
- **An unknown organisation is a web page, not a 404.** The API answers a name
  it doesn't know with the single-page app's own `index.html` and a 200, so a
  mistyped org looked exactly like a corrupt response. It now says what it is,
  including that the name is case-sensitive — `StackOne` works, `stackone` gets
  the web page, and nothing else would tell you.
- **Colours are read loosely and written strictly.** The host requires a leading
  hash and drops a bare `ff0000` to the accent colour, so every colour the
  document carries is normalised on the way out.

## Configuring one

The Derby needs an organisation before it can draw anything, and the only way
to set one was to hand-edit `~/.stack-nudge/config`.

A manifest's `config` entry may now be a bare key name **or** an object carrying
a label, help text and a placeholder. One list rather than two: a parallel array
describing the keys would drift from the list actually passed, and what you
would get is a form field for a key nobody reads, or a key nobody can set.

The published index deliberately carries key *names* only. It is a wire format
read by binaries of every version — the app fetches it from `/releases/latest`,
so an old host reads the newest index — and the form never needs the metadata
there, since it renders for an extension that is already installed and reads
that extension's own manifest.

## Finding one

**Settings → Extensions** lists a card per installed extension, then
**Browse extensions…**. What you have installed is a setting; the browser is for
finding what you don't.

Cards rather than the flat label/value row every other category uses, because a
card already carries version, description, the keys it reads and what it needs.
It is still a `SettingsRow`, so the selection ring, scroll-to-index and arrow
keys work on it unchanged.

Every installed extension opens a page, including one declaring no config keys:
the page is where Remove lives, and a row that opened nothing would make Enter
mean something different depending on the extension. That also takes Remove off
the browser, where Enter previously meant install, update, dismiss *or* delete
depending on which row the selection happened to be on — and the destructive end
of that is the one you hit by accident.

Refusals stay with the installed extensions. A refusal is an installed thing
that is broken, and splitting the two apart is what produced an orange "Remove"
sitting directly above a card offering "Install" in the first place.

## Things worth a reviewer's eye

**A manifest could write any config key.** The namespace guard tested whether a
key *started* in `STACKNUDGE_EXT_`, which was enough while keys were only read.
The form inverted that: a key declared as
`"STACKNUDGE_EXT_DERBY_ORG\nSTACKNUDGE_CLAUDE_PATH=/tmp/evil\n#"` rendered a
labelled "Organisation" field, and saving it wrote three lines — the middle one
setting the CLI path that `ProcessOutput.claude()` executes and `notify.sh`
sources. The injected line outlived the extension, so it defeated the review
that is meant to be the control and persisted past the removal that is meant to
be the remedy. Closed at both ends, and the end anchor is now `\z` rather than
`$`, which is safe under `NSRegularExpression` as well.

**The index format was a breaking change.** Object-form `config` in
`extensions-index.json` would have failed the *whole document* for any host with
the previous `[String]` decoder — every entry, not just the derby's. Entries
decode leniently now, so a future entry-shape change degrades rather than breaks.

**Keyboard routing on the browser was rewritten twice.** The first attempt
focused the search field on arrival and stopped swallowing unrecognised keys so
letters could "reach SwiftUI" — backwards, since a focused field is first
responder and `FloatingPanel.keyDown` only fires for what it declines. Letters
never reached the handler; what changed was that Esc, ↑↓ and ⏎ stopped reaching
it while the footer advertised them. It now copies the history pane's contract
exactly, including the field-editor selection collapse that pane needed — without
it, typing `de` gave you `e`.

**`R` is now `⌘R`** on the browser, since a plain letter is a search term.

**`PRIVACY.md` said the app "has no network access".** Untrue for some time —
update checks, the extension index, Slack — and shipping an extension that GETs
a third-party race made it worse rather than stale. It now says what goes out and
when, including that the `gh` fallback authenticates as you, and that the voice
model is ~350 MB in `~/.cache/stackvox/`, outside the directory the removal
instructions named.

**A packaging bug:** the packer swept `__pycache__` into release tarballs, so an
extension with tests shipped stale bytecode and the archive's bytes depended on
whether the builder had run them.

## Verification

- 73 Python tests for the extension and 1134 Swift tests, both run by CI. A
  suite that discovers zero tests is a failure rather than a pass, and the
  Python suite fails loudly if anything reaches the network — patching the wrong
  seam once let it quietly make live requests, and the only symptom was five
  seconds instead of two milliseconds.
- Mutation-verified in dedicated worktrees across three rounds.
- Five adversarial review agents, which found the config-key injection, the
  index break, the keyboard regression and the field-editor bug. All four are
  things I would have shipped.
- Checked end to end against the live service throughout.

---

## Release

`Release-As: 2.0.0`

Cut as **2.0.0** rather than 1.36.0. The PR title carries `!`, so release-please
bumps the major from the squash commit's subject — the `Release-As` footer above
is a backstop in case the squash message is edited by hand, since this repo's
squash bodies are a list of commit subjects and drop commit bodies.

What changes for someone already on 1.35.0, which shipped the extensions
browser:

- `R` no longer reloads the catalogue; it types into the search field. Reload is
  `⌘R`.
- `⏎` on an installed extension no longer removes it. It opens that extension's
  page, where Remove lives. Removing by accident was a keystroke away before.
- Installed extensions are managed in **Settings → Extensions**, not inside the
  browser. The browser is for finding extensions you don't have.
- `Space` is no longer an activation key on the browser; it is a space.


extensions/derby/run
```diff
@@ -0,0 +1,585 @@
...
+    # Deduplicated here rather than left to the host, which drops the second of
+    # a duplicate pair without saying so. str() also collides 1 with "1", and
+    # the id is what an action names a row by — two rows answering to one id is
+    # a row that acts on the wrong horse.
+    horses = []
+    seen_ids = set()
+    for entry in race.get("horses") or []:
+        if not isinstance(entry, dict):
+            continue
+        identifier = text(entry.get("horse_id")) or str(entry.get("horse_id") or "")
```
StuBehan: what happens to a horse with `horse_id: 0`? `or ""` makes it falsy so it hits the `not identifier` guard and the whole row goes, and because `leader` is a max over whatever survived, the next horse gets promoted to the finish line. `False` does the same. Ran it:

```
horses: [{horse_id: 0, name: "Zero", tokens: 100}, {horse_id: 1, name: "One", tokens: 50}]
rows:   [("1", "One")]   state: "ok"
```

`is None` rather than truthiness? Also the identifier here is trimmed and line 338 emits the untrimmed `str()`, so `" 1 "` dedupes as `1` but renders as `" 1 "`. Worth deriving once.

extensions/derby/run
```diff
@@ -0,0 +1,585 @@
...
+import sys
+import urllib.error
+import urllib.parse
+import urllib.request
+from datetime import datetime, timezone
+
+DEFAULT_BASE = "https://token-derby.mauricode.co.uk/api"
+ORG_KEY = "STACKNUDGE_EXT_DERBY_ORG"
+BASE_KEY = "STACKNUDGE_EXT_DERBY_BASE"
+TIMEOUT = 8  # The host's own budget is 12s for two round trips plus overhead.
```
StuBehan: the comment says the host budget is 12s for two round trips, but this is 8s per request and there are two sequential hops, so two slow-but-successful ones is 16s and `ExtensionRuntime.timeout` kills the process before the script gets to print its own error. 5s?

panel/ExtensionConfigView.swift
```diff
@@ -0,0 +1,278 @@
...
+                FooterHint(label: "Back", keys: ["Esc"])
+                // Only where there is something to save. This page now opens
+                // for every installed extension, including ones declaring no
+                // keys — and a refused one, which is the most common case of
+                // all — so an unconditional Save hint promised a key that had
+                // no button and no handler behind it.
+                if !model.keys.isEmpty {
+                    FooterHint(label: "Save", keys: ["⏎"])
+                }
+                FooterHint(label: "Remove", keys: ["⌘⌫"])
```
StuBehan, line 193: ⌘⌫ is a standard field editor binding (`deleteToBeginningOfLine`) so it gets consumed before `FloatingPanel.keyDown` ever sees it, which means this hint stops working the second anyone clicks into a field. Esc and ⏎ both got a field level handler for exactly that reason. Should this get one too, or hide the hint while the field has focus?

extensions/derby/run
```diff
@@ -0,0 +1,585 @@
...
+    """https only, and no credentials in the URL.
+
+    The value comes from the config file, which is the user's own, so this is
+    not a trust boundary — but a base silently downgraded to http would send a
+    request the user thinks is encrypted, and that is worth refusing loudly.
+    """
+    if not value:
+        return DEFAULT_BASE
+    parsed = urllib.parse.urlsplit(value)
+    if parsed.scheme != "https" or not parsed.netloc or parsed.username:
```
StuBehan: docstring says no credentials in the URL, but `parsed.username` is `''` when the userinfo is just `:pw`, and `''` is falsy:

```
https://user:pw@host/api  -> None
https://:pw@host/api      -> 'https://:pw@host/api'
```

`or parsed.password` as well?

panel/PanelNav.swift
```diff
@@ -920,6 +975,33 @@ final class PanelNav: ObservableObject {
...
+        }
+    }
+    // The form behind .extensionConfig. Held here rather than built in the view
+    // so its unsaved edits survive a re-render, and cleared on the way out so a
+    // second visit reads the file again rather than showing the last visit's
+    // values.
+    @Published var extensionConfig: ExtensionConfigModel?
 
     @Published var extensionTabs: [ExtensionTab] = [] {
         didSet { reconcileModeWithTabs() }
```
StuBehan, line 1007: `reconcileModeWithTabs` only guards `.extensionTab` and there's a mode a level below it now. Its own comment says an extension removed while its tab is open leaves mode on something that renders nothing, which is the same thing that happens to `.extensionConfig(id)` if the directory goes away out of band. `removeExtension` covers the in-app path so its only really reachable by hand, but should this catch both?

panel/ExtensionInstaller.swift
```diff
@@ -69,20 +77,35 @@ enum ExtensionInstaller {
...
-            let extensions: [IndexEntry]?
+            let extensions: [Lenient]?
         }
         let document: Document
         do {
             document = try JSONDecoder().decode(Document.self, from: data)
         } catch {
             return .failure(.malformedIndex("not valid JSON"))
         }
+        // Note: entries decode leniently (see Lenient below), so one entry this
```
StuBehan: `Lenient` is 16 lines above this, not below, and this repeats what its own comment already says. Its also sat on the schema guard, which isn't about leniency at all. Not blocking but worth a tidy.

extensions/derby/run
```diff
@@ -0,0 +1,585 @@
+#!/usr/bin/env python3
```
StuBehan, line 1: counted this at 43% prose against 30% for `extensions/system/run` and 25% across `panel/`. The facts in the docstrings are all worth having, its the narrative framing round them I'd bin off. `pick()` and `number()` especially read as three paragraphs where one sentence does it.

extensions/derby/test_derby.py
```diff
@@ -0,0 +1,691 @@
...
+    # The field is the service's, and a non-string there put a JSON object
+    # where the host's decoder wants a string — which fails the *whole*
+    # document, so the pane would say "wrong type for message" instead of
+    # showing the API's error.
+    def test_a_non_string_error_message_does_not_reach_the_document(self):
+        import io
+        for body in ['{"code":"X","message":{"a":1}}', '{"code":"X","message":7}',
+                     '{"code":"X"}', 'not json']:
+            derby._opener = self.raising_stub(
+                lambda url, _b=body: derby.urllib.error.HTTPError(
```
StuBehan, line 681: suite leaks these into CI output, three of them:

```
ResourceWarning: Implicitly cleaning up <HTTPError 404: 'no'>
```

`HTTPError` fixtures aren't being closed. Cosmetic, but it makes a green run look like it isn't one.

StuBehan (changes requested):
The config key fix is solid, and closing it at the write site as well was the right call. The privacy policy merge is right too, taking #192's structure and adding what the extensions do on top. Rest is inline.

extensions/derby/run
hiskudin: Worse than that — I reproduced it and the promotion is visible: with horses 0 (100 tokens) and 1 (50), the row for 0 vanished and 1 rendered at fill 0.5, i.e. leading, because `leader` is a max over the survivors. So the wrong horse appears to be winning rather than just one going missing.

Fixed by deriving the id once in a `row_id()` helper that returns `None` only when `horse_id` is genuinely absent, and using that same value for both the dedupe key and the emitted `id`. That closes the `" 1 "` inconsistency you spotted in the same breath. Tests for a falsy id keeping its row and its lead, for a missing id being dropped, and for the deduped id being the emitted one.

extensions/derby/run
hiskudin: Right, and the failure mode is the bad one: the host kills the process, so the script never gets to print its own `error` document and a slow service reads as a crashed extension rather than a slow one.

Taken your number — 5s, so two hops plus overhead fit inside 12s. Added a test asserting `TIMEOUT * 2 < 12` so the relationship is pinned rather than left to the comment.

panel/ExtensionConfigView.swift
hiskudin: Good catch, and it is the third time in this PR that a focused field editor has quietly eaten a binding — the same mechanism that broke Esc and ↑↓ on the browser.

Gone with dimming rather than a field-level handler. There is no SwiftUI hook for ⌘⌫ the way `.onExitCommand` and `.onSubmit` exist for Esc and Return, so any handler would mean reaching for the responder chain to do what the field editor is correctly already doing. The hint now dims while a field has focus, which is the treatment the Settings footer already gives its Cycle hint on rows where the arrows do nothing. Most pages here have no fields at all — `system` and every refused extension — so ⌘⌫ works on them unconditionally.

extensions/derby/run
hiskudin: Confirmed — `https://:pw@host/api` came back as itself. Checking `parsed.password` as well now, and both that and a bare `user@` are in the rejection list.

panel/PanelNav.swift
hiskudin: Yes, both. `removeExtension` covers the in-app path, so what is left is the directory going away underneath the panel — and that leaves the mode on a page describing something that no longer exists, which is exactly what the tab guard exists to prevent one level up.

`reconcileModeWithTabs` now sends `.extensionConfig(id)` back to Settings when the id is no longer installed, checked before the tab guard. Test covers both halves: the page survives while the extension does, and closes when it does not.

panel/ExtensionInstaller.swift
hiskudin: Fair — it restated the comment on `Lenient` and sat on the schema guard, which is unrelated. Removed; the reasoning stays in one place on the type itself.

extensions/derby/run
hiskudin: Agreed, and thanks for measuring it rather than just asserting it.

Cut the module docstring to what it is and what it talks to, and took the narrative framing out of `pick`, `number`, `position`, `scored`, `normalise_hex`, `parse_hex`, `mane_for`, `NotJSON` and `UnknownOrg` — mostly by deleting the paragraph that explained why the first version was wrong, which belongs in the commit rather than the file. The facts stayed, per your note. Roughly 35 lines of prose gone.

extensions/derby/test_derby.py
hiskudin: Fixed. `HTTPError` wraps an open file object, so the fixtures are collected and closed in `tearDown` now. Verified under `-W error::ResourceWarning` with a forced `gc.collect()` — clean.

